PR #4, opened 2023-05-08
Feature/srwopschall/mortar naming

### LINKED ISSUE #1: Remove master/slave terminology
There are classes that still uses this. It should be renamed.

srwopschall:
Per discussion with @ebchin 
Changed master -> mortar
Changed slave -> nonmortar


ebchin:
Thanks @srwopschall!  It might be a good idea to roll changing 1->mortar and 2->nonmortar in this PR so we can eliminate the 1/2 numbering as well.

srwopschall:
> Thanks @srwopschall! It might be a good idea to roll changing 1->mortar and 2->nonmortar in this PR so we can eliminate the 1/2 numbering as well.

@ebchin - where specifically are you referring to in the code? We have many situations where we have side 1 and side 2, but in mortar specific context we translate 1->mortar and 2->nonmortar. We don't want this to propagate everywhere because it isn't relevant for common plane. This is slightly tricky, because the test utility class was developed to support testing the mortar, but common plane tests use this, so you see the lexicon trickle into those tests via the TestMesh class, but this is localized.

ebchin:
> @ebchin - where specifically are you referring to in the code? We have many situations where we have side 1 and side 2, but in mortar specific context we translate 1->mortar and 2->nonmortar. We don't want this to propagate everywhere because it isn't relevant for common plane. This is slightly tricky, because the test utility class was developed to support testing the mortar, but common plane tests use this, so you see the lexicon trickle into those tests via the TestMesh class, but this is localized.

Yeah, that makes sense re:TestMesh; I didn't realize these data structures are used for common plane as well. I do see a few 1/2 references in [Mortar.cpp](https://github.com/LLNL/Tribol/blob/ac681356f25a4d14b1aacd434a90d4281d20ccc4/src/tribol/physics/Mortar.cpp#L352) using Mortar ->1, Nonmortar -> 2 numbering, but this can be handled in a future PR.

ebchin (approved):

PR #5, opened 2023-05-13
Update umpire and fix bug in tests

ebchin:
This PR fixes failing tests when umpire is enabled.  Specifically, it updates the version of Umpire used in the spack package and fixes a bug when `umpire::ResourceManager` initializes after `tribol::CouplingSchemeManager`.


scripts/spack/packages/tribol/package.py
```diff
@@ -83,7 +83,7 @@ class Tribol(CachedCMakePackage, CudaPackage):
     depends_on("raja", when="+raja")
     depends_on("raja~openmp~shared~examples~exercises", when="+raja+minbuild")
 
-    depends_on("umpire@6.0.0tribol", when="+umpire")
+    depends_on("umpire@2022.03.1", when="+umpire")
```
white238:
```suggestion
    depends_on("umpire@2022.03.1:", when="+umpire")
```

white238 (approved):

scripts/spack/packages/umpire/package.py
```diff
@@ -21,12 +21,9 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     maintainers = ['davidbeckingsale']
 
-    # TRIBOL EDIT START
-    version('6.0.0tribol', commit='5f474c7501daf365d6015a2141821d83bd799ffa', submodules="True")
```
srwopschall, line 25: I'm assuming there is no real reason to do it this way, where we have our own tribol version pointing to some commit sha? And per below we can just point to a version of umpire?

scripts/spack/packages/umpire/package.py
```diff
@@ -58,13 +55,10 @@ class Umpire(CachedCMakePackage, CudaPackage, ROCmPackage):
...
     patch('missing_header_for_numeric_limits.patch', when='@4.1:5.0.1')
-    # TRIBOL EDIT BEGIN
-    patch('cmake_devicealloc_option_6.0.0serac.patch', when='@6.0.0tribol')
-    # TRIBOL EDIT END
 
     # export targets when building pre-6.0.0 release with BLT 0.4.0+
-    patch('https://github.com/LLNL/Umpire/commit/5773ce9af88952c8d23f9bcdcb2e503ceda40763.patch',
-          sha256='f5c691752e4833a936bce224bbe0fe884d3afa84c5e5a4a481f59a12840159c9',
+    patch('https://github.com/LLNL/Umpire/commit/5773ce9af88952c8d23f9bcdcb2e503ceda40763.patch?full_index=1',
+          sha256='f3b21335ce5cf9c0fecc852a94dfec90fb5703032ac97f9fee104af9408d8899',
```
srwopschall, line 61: I don't know the background pertaining to this or the particular sha, but I trust this is correct.

scripts/spack/packages/umpire/package.py
```diff
@@ -183,6 +184,17 @@ def initconfig_hardware_entries(self):
                 arch_str = ",".join(archs)
                 entries.append(cmake_cache_string(
                     "HIP_HIPCC_FLAGS", '--amdgpu-target={0}'.format(arch_str)))
+            # BEGIN SERAC EDIT
```
srwopschall: We may want to take out references to others codes as a general practice? Can we adopt this as a # TRIBOL EDIT?

scripts/spack/packages/umpire/package.py
```diff
@@ -200,18 +214,22 @@ def initconfig_package_entries(self):
...
+            "{}ENABLE_EXAMPLES".format(option_prefix), '+examples' in spec))
+        entries.append(cmake_cache_option(
+            "{}ENABLE_DOCS".format(option_prefix), False))
         entries.append(cmake_cache_option("BUILD_SHARED_LIBS", '+shared' in spec))
         entries.append(cmake_cache_option("ENABLE_TESTS", 'tests=none' not in spec))
-
-        # TRIBOL EDIT BEGIN
-        entries.append(cmake_cache_option("UMPIRE_ENABLE_DEVICE_ALLOCATOR", '+device_alloc' in spec))
-        # TRIBOL EDIT END
+    # BEGIN SERAC EDIT
```
srwopschall: Ref to SERAC?

srwopschall (approved):
Minor comment about whether to keep reference to SERAC in here.

scripts/spack/packages/tribol/package.py
ebchin: Thanks for the suggestion.  Fixed in the latest commit.

scripts/spack/packages/umpire/package.py
ebchin: The previous commit sha is based on what was in Serac at the time and the updated version is based on what's in Serac now.

scripts/spack/packages/umpire/package.py
ebchin: I grabbed it from Serac's package.py from Umpire and looks like it matches what's in the spack repo.  See [here](https://github.com/LLNL/serac/blob/develop/scripts/spack/packages/umpire/package.py) and [here](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/umpire/package.py).

scripts/spack/packages/umpire/package.py
ebchin: Thanks for catching this.  Fixed in the last commit.

scripts/spack/packages/umpire/package.py
ebchin: Fixed

PR #8, opened 2023-05-31
Feature/srwopschall/testmesh_tets

srwopschall:
Adds ability to create tet mesh decomposition of hex mesh on the TestMesh class.
Adds ability to convert tet "test mesh" to mfem tet mesh. 
Added tests for the tet TestMesh construction and the mfem tet mesh construction.
Adding finite element support for tet meshes and updating examples, tests, and methods appropriately will be in follow on PR.


srwopschall:
@ebchin - I fixed the tribol_coupling_scheme test and all checks have passed. Memory management was ok. I had to update some element type modifications I had made elsewhere.

src/tests/tribol_tet_mesh.cpp
```diff
@@ -0,0 +1,177 @@
...
+   EXPECT_EQ( this->m_mesh.mfem_mesh->GetNE(), this->m_mesh.numMortarElements + this->m_mesh.numNonmortarElements );
+   EXPECT_EQ( this->m_mesh.mfem_mesh->GetNV(), this->m_mesh.numMortarNodes + this->m_mesh.numNonmortarNodes );
+   EXPECT_EQ( this->m_mesh.mfem_mesh->GetNFbyType(mfem::FaceType::Boundary), 
+              2 * 6 * (nMortarElems*nMortarElems + nNonmortarElems*nNonmortarElems) );
+
+   // check for inverted elements
+   int wrong_elem_orientation = -1;
+   int wrong_bdr_elem_orientation = -1;
+   bool fix_it = false;
+   wrong_elem_orientation = this->m_mesh.mfem_mesh->CheckElementOrientation( fix_it );
```
ebchin, line 161: I don't think this will test the orientation of the Tribol mesh, since when the mfem mesh is finalized during the call to `TestMesh::setupMfemMesh()`, the orientations will be fixed.  You can call `FinalizeTetMesh()` with `fix_orientation` false, but...

src/tests/tribol_tet_mesh.cpp
```diff
@@ -0,0 +1,177 @@
...
+   wrong_bdr_elem_orientation = this->m_mesh.mfem_mesh->CheckBdrElementOrientation( fix_it );
```
ebchin, line 162: ...it looks like it won't fix this call, since calling `FinalizeTetMesh()` will [always fix boundary orientations](https://github.com/mfem/mfem/blob/master/mesh/mesh.hpp#L810).

src/tribol/common/Parameters.hpp
```diff
@@ -36,11 +36,16 @@ constexpr integer ANY_MESH = -1;
...
-   NODE,              ///! 0D node
-   EDGE,              ///! 1D edge
-   FACE,              ///! 2D face
-   CELL,              ///! 3D volume
-   NUM_CONTACT_ELEMENTS = CELL
+//   NODE,              ///! 0D node
+//   EDGE,              ///! 1D edge
+//   FACE,              ///! 2D face
+//   CELL,              ///! 3D volume
+   LINEAR_EDGE,       ///! 1D linear edge
```
ebchin, line 43: It would be better to take the element order out of this enum so it doesn't define multiple things at once.

src/tribol/physics/AlignedMortar.cpp
```diff
@@ -209,6 +208,7 @@ void ComputeAlignedMortarGaps( CouplingScheme const * cs )
 
    MeshData& mortarMesh = meshManager.GetMeshInstance( mortarId );
    MeshData& nonmortarMesh  = meshManager.GetMeshInstance( nonmortarId );
+   IndexType const numNodesPerFace = mortarMesh.m_numCellNodes;
```
ebchin: `m_numNodesPerCell` might be clearer than `m_numCellNodes`.

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -53,7 +53,8 @@ bool geomFilter( InterfacePair const & iPair, ContactMode const mode )
    MeshData& mesh1 = meshManager.GetMeshInstance(meshId1);
    MeshData& mesh2 = meshManager.GetMeshInstance(meshId2);
 
-   int dim = ( mesh1.m_elementType == tribol::EDGE ) ? 2 : 3;
+//   int dim = ( mesh1.m_elementType == tribol::EDGE ) ? 2 : 3;
```
ebchin: Why is this line commented out?  Does it still need to be here?

src/tribol/utils/TestUtils.cpp
```diff
@@ -94,28 +94,32 @@ TestMesh::~TestMesh()
 }
 
 //------------------------------------------------------------------------------
-void TestMesh::clear()
+void TestMesh::clear( bool keepCoords )
 {
-   if (this->mfem_mesh != nullptr)
```
ebchin, line 99: I'm thinking we should keep `this->` to be consistent with the way we address member variables in this method.  More generally, it might be good to have a project-wide convention for naming and addressing member variables, parameters, etc.

src/tribol/utils/TestUtils.cpp
```diff
@@ -363,37 +371,29 @@ void TestMesh::setupContactMeshHex( int numElemsX1, int numElemsY1, int numElems
...
+   this->elConn1   = new int[ this->numNodesPerElement * this->numMortarElements ];
+   this->elConn2   = new int[ this->numNodesPerElement * this->numNonmortarElements ];
    this->faceConn1 = new int[ this->numNodesPerFace * this->numMortarFaces ];
    this->faceConn2 = new int[ this->numNodesPerFace * this->numNonmortarFaces ];
-   this->x = new real[ this->numTotalNodes ];
-   this->y = new real[ this->numTotalNodes ];
-   this->z = new real[ this->numTotalNodes ];
+   this->x = new double[ this->numTotalNodes ];
+   this->y = new double[ this->numTotalNodes ];
+   this->z = new double[ this->numTotalNodes ];
```
ebchin, line 396: Why is this a `double` vs a `real`?  If the plan is to get rid of tribol's data types in favor of the standard C++ ones, that's fine (I'm not sure if there's a reason to support single precision `float`?).   It would be good to standardize this throughout the codebase though (except when non-standard types like `axom::IndexType` or `HYPRE_BigInt` are called for).

src/tribol/utils/TestUtils.cpp
```diff
@@ -615,24 +827,20 @@ void TestMesh::allocateAndSetVelocities( int meshId, real valX, real valY, real
...
-   // check mesh ids. While Tribol supports auto-contact, the TestMesh class 
-   // has to have unique mesh Ids set prior to calling this routine
-   if (mortarMeshId == nonmortarMeshId)
-   {
-      SLIC_WARNING("TestMesh::allocateAndSetVelocities(): please set unique " << 
-                   "mortarMeshId and nonmortarMeshId prior to calling this routine.");
-   }
+   // Check that mesh ids are the same. The TestMesh class was built around 
+   // testing the mortar method with Lagrange multiplier enforcement, which does 
+   // not support auto contact.
```
ebchin, line 846: 👍

src/tribol/utils/TestUtils.cpp
```diff
@@ -1015,17 +1216,29 @@ int TestMesh::tribolSetupAndUpdate( ContactMethod method,
...
-                                           bool mortar, 
                                            int nodeIdOffset, 
                                            bool inHomogeneousGap, 
                                            real inHomogeneousZVal )
 {
+   SLIC_ERROR_IF( !this->mesh_constructed, "TestMesh::setupPatchTestDirichletBCs(): " << 
+                  "mesh must be constructed prior to calling this routine." );
+
+   bool mortar = false;
+   if (meshId == mortarMeshId)
```
ebchin: `this->mortarMeshId` (I'll stop commenting on these...)

src/tribol/utils/TestUtils.cpp
```diff
@@ -546,18 +546,230 @@ void TestMesh::setupContactMeshHex( int numElemsX1, int numElemsY1, int numElems
...
+   this->mesh_constructed = true;
+
+} // end setupContactMeshTet()
+//------------------------------------------------------------------------------
+void TestMesh::allocateAndSetVelocities( int meshId, real valX, real valY, real valZ )
+{
+   // Check that mesh ids are not the same. The TestMesh class was built around 
+   // testing the mortar method with Lagrange multiplier enforcement, which does not 
+   // support auto contact.
+   SLIC_WARNING_IF( mortarMeshId == nonmortarMeshId,
```
ebchin: `this->mortarMeshId`

src/tribol/utils/TestUtils.cpp
```diff
@@ -615,24 +827,20 @@ void TestMesh::allocateAndSetVelocities( int meshId, real valX, real valY, real
...
-   // has to have unique mesh Ids set prior to calling this routine
-   if (mortarMeshId == nonmortarMeshId)
-   {
-      SLIC_WARNING("TestMesh::allocateAndSetVelocities(): please set unique " << 
-                   "mortarMeshId and nonmortarMeshId prior to calling this routine.");
-   }
+   // Check that mesh ids are the same. The TestMesh class was built around 
+   // testing the mortar method with Lagrange multiplier enforcement, which does 
+   // not support auto contact.
+   SLIC_WARNING_IF( mortarMeshId == nonmortarMeshId,
```
ebchin: `this->mortarMeshId`, `this->nonmortarMeshId`

src/tribol/utils/TestUtils.cpp
```diff
@@ -847,24 +1040,32 @@ int TestMesh::tribolSetupAndUpdate( ContactMethod method,
...
+         allocRealArray( &this->gaps, this->numTotalNodes, 0. );
+         allocRealArray( &this->pressures, this->numTotalNodes, 1. );
 
-   }
-   else if (method == MORTAR_WEIGHTS)
-   {
-      allocRealArray( &this->gaps, this->numTotalNodes, 0. );
-      this->pressures = nullptr;
+         // register nodal gaps and pressures
+         registerRealNodalField( nonmortarMeshId, MORTAR_GAPS, this->gaps );
```
ebchin: `this->nonmortarMeshId`

ebchin (commented):
Thanks @srwopschall, I had some minor comments but it looks good overall.  Looking forward to seeing the upcoming PRs.

src/tests/tribol_tet_mesh.cpp
srwopschall: I see your point here. Even if my TestMesh has inverted elements, let's say, or inverted surface unit normals, `FinalizeTetMesh()` will fix these so this isn't really testing anything. I'll think of another way to test the validity of the mesh. Not sure there is actually a good test for the mfem created mesh, since orientation issues will be fixed. @ebchin - suggestions?

srwopschall: Orientation and checks have been updated.

src/tribol/common/Parameters.hpp
srwopschall: This was the part that will likely change when we go to high order, so this is an intermediate step. The order + topology serves only to tell me how many nodes there are. The previous EDGE, FACE, CELL isn't enough because a Triangle and a quadrilateral are both faces. If I use this to tell how many nodes a face has, then I assume (for now) they are linear faces. Otherwise, I'd have to have another enum for the order. Here, I baked all that into one for ease. I think going to high-order will be a big enough refactor that no matter what I do here, we will end up touching and modifying it. Thoughts @ebchin ?

src/tribol/physics/AlignedMortar.cpp
srwopschall: Fixed. Good call. This is legacy naming. Your suggestion makes more sense.

src/tribol/search/InterfacePairFinder.cpp
srwopschall: Removed. Doesn't need to be there.

src/tribol/utils/TestUtils.cpp
srwopschall: Yes. I think that could be a straightforward refactor. The convention I'm used to is that member variables are m_<name>, and when you see that you don't need this->. I've probably used this-> too much, but for this class there is no 'm_' so you can't easily tell. Let's refactor that in another cleanup PR.

src/tribol/utils/TestUtils.cpp
srwopschall: This is likely an old convention that I have mangled. I'll create an issue and we can do a quick PR.

src/tribol/utils/TestUtils.cpp
srwopschall: Fixed.

src/tribol/utils/TestUtils.cpp
srwopschall: Fixed.

src/tribol/utils/TestUtils.cpp
srwopschall: Fixed. Good catches.

src/tribol/common/Parameters.hpp
ebchin: Makes sense to me, I'm fine with leaving it as is.  This enum uniquely defines an element type instead of an element shape.

ebchin (approved):

PR #9, opened 2023-06-01
Gitlab CI

white238:
Gitlab CI passed on Quartz here:

https://lc.llnl.gov/gitlab/tribol/tribol/-/pipelines/293524

As soon as this is run on the default github branch, `develop`, I can enable it to be a required PR CI check.


.gitlab/build_quartz.yml
```diff
@@ -0,0 +1,52 @@
...
+  variables:
+    ALLOC_OPTIONS: "-p pdebug --exclusive --mpibind=off"
+    ALLOC_COMMAND: "salloc ${ALLOC_OPTIONS} -N ${ALLOC_NODES} -t ${ALLOC_TIME}"
+  tags:
+    - shell
+    - quartz
+  rules:
+    - if: '$CI_COMMIT_BRANCH =~ /_qnone/ || $ON_QUARTZ == "OFF"' #run except if ...
+      when: never
+    - when: on_success
```
adrienbernede, line 20: When a job has a `rules` section, the parsing will stop at the first conditional that resolves to "true" and apply the associated `when` clause.
Illustration:
- The above rules section means:
  1. If the branch name contains "_qnone" or the CI variable "ON_QUARTZ" is set to "OFF", then the job will never run. (That a way to run your pipeline with no access to quartz, for testing, for days when quartz is offline, or for a personal clone by an intern with partial access rights).
  2. else, run if previous jobs passed.
  
- No conditional resolves to "true" (see that has an "else"). If the parsing reaches a rule with no conditional, the subsequent rules will never be parsed. The following code means the job will run if the status of the required preceding jobs is success.
```
job:
  rules:
    - /* else */ when: on_success
    # it’s useless to add other rules here.
```
- When no rule section is provided, it’s equivalent to that last example: having a rule with `when: on_success`. Gitlab provides a shortcut to change this default: you can have the when clause directly in the job:
```
job:
  when: on_failure
```
- In a `rules` section, If no "true" statement is found, the job will not run. It’s like if there was a last rule: "-when: never".
- In Serac, rules are split in two job templates. GitLab will stack them depending on the order of definition (the order in which the templates where defined in the job, see the `extends` section). This is not recommended as this is brittle.
Instead, I recommend setting all the rules for all the jobs at once. An example in the radiuss-shared-ci implementation:
https://github.com/LLNL/radiuss-shared-ci/blob/c9c37efcd7eecfbdc99388ad8063daf4097bf788/ruby-build-and-test.yml#L16

host-configs/toss_4_x86_64_ib-clang@14.0.6.cmake
```diff
@@ -0,0 +1,114 @@
...
+set(ENABLE_MPI ON CACHE BOOL "")
+
+#------------------------------------------------------------------------------
+# Hardware
+#------------------------------------------------------------------------------
+
+set(ENABLE_OPENMP ON CACHE BOOL "")
+
+#------------------------------------------------------------------------------
+# TPLs
```
srwopschall, line 53: Do we need all of these TPLs for the Tribol project? Any advantage/clarity gained from reducing this list?

host-configs/toss_4_x86_64_ib-gcc@10.3.1.cmake
```diff
@@ -0,0 +1,114 @@
...
+set(ENABLE_MPI ON CACHE BOOL "")
+
+#------------------------------------------------------------------------------
+# Hardware
+#------------------------------------------------------------------------------
+
+set(ENABLE_OPENMP ON CACHE BOOL "")
+
+#------------------------------------------------------------------------------
+# TPLs
```
srwopschall, line 53: Same TPL comment as in clang host-config.

srwopschall (commented):
Thanks @white238 - this will be great to have. I defer ultimate approval to reviewers more familiar with standing up CI (1 reviewer approval required to merge).

host-configs/toss_4_x86_64_ib-clang@14.0.6.cmake
white238: \<whine\>FINE I won't be lazy and just copy/paste....\</whine\>

I'll delete some stuff later today.

.gitlab/build_quartz.yml
```diff
@@ -0,0 +1,52 @@
...
+    - module load python/3.9.12
+    # We could also extract the CMake executable location from the hostconfig in common_build_functions
+    # like we do in config-build
+    - module load cmake/3.19.2
+
+
+####
+# Templates
+.src_build_on_quartz:
+  extends: [.src_build_script, .on_quartz]
```
adrienbernede, line 31: Maybe a comment stating that because of variables nesting to define the allocation command, the order in the extends list matters.

adrienbernede (approved):
The CI is working, the design is known to work in Serac, and it’s extensible to other machines. Let’s move forward ;)

adrienbernede (approved):

adrienbernede:
I don’t have write access, so my approval does not count...

srwopschall (approved):
Approving based on Adrien's approval; working around write access issue.

PR #10, opened 2023-06-02
fixed node value summation in GalerkinEval()

ebchin:
Sums up the correct node values with the basis function value.


srwopschall (approved):
@ebchin - good catch. Thank you.

PR #16, opened 2023-06-26
Bugfix/srwopschall/valid velocity

srwopschall:
- Allows the registration of null nodal velocity and nodal response pointers in certain cases.
- Moves check for valid velocity pointers (null vs. non-null) to coupling scheme and mesh level.
- Coupling scheme tests added
- Misc. cleanup


src/tribol/interface/tribol.cpp
```diff
@@ -319,21 +319,26 @@ void registerMesh( integer meshId,
...
-   if (x == nullptr || y == nullptr)
-   {
-      SLIC_WARNING("Pointer to x and y-component mesh coordinate arrays are null pointers " <<
-                   " for mesh id, " << meshId << ".");
-      mesh.m_isValid = false;
-   }
 
-   if (dim == 3)
+   // check for null pointers for non-null meshes
+   if (numCells > 0)
```
liu15, line 324: How is numCells determined?

src/tribol/interface/tribol.cpp
```diff
@@ -319,21 +319,26 @@ void registerMesh( integer meshId,
...
    {
-      if (z == nullptr)
+      if (x == nullptr || y == nullptr)
       {
-         SLIC_WARNING("Pointer to z-component mesh coordinates is null for " << 
-                      "mesh id, " << meshId << ".");
+         SLIC_WARNING("Pointer to x and y-component mesh coordinate arrays are null pointers " <<
```
liu15: should probably be "x or y"

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1158,11 +1148,8 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
    KinematicPenaltyCalculation kin_calc = pen_enfrc_options.kinematic_calculation;
    if ( kin_calc == KINEMATIC_CONSTANT )
    {
-      SLIC_WARNING("Tribol timestep vote may be inaccurate when " << 
-                   "using constant kinematic penalty option with " << 
-                   "penalty enforced methods. Consider registering " << 
-                   "'element_wise' data in call to tribol::setKinematicElementPenalty() " << 
-                   "for element-specific penalty calculations.");
+      // Tribol timestep vote only used with KINEMATIC_ELEMENT penalty is used
+      // because element thicknesses are supplied
```
liu15, line 1167: This comment doesn't make sense

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1063,35 +1074,23 @@ void CouplingScheme::computeTimeStep(real &dt)
       if (mesh1.m_velX == nullptr || mesh1.m_velY == nullptr)
       {
          meshVel1 = false;
-         SLIC_WARNING( "CouplingScheme::computeTimeStep(): please register nodal "  << 
-                       "velocities for mesh id, " << m_meshId1 << ", in order to " <<
```
liu15, line 1067: Would it make sense to return an invalid value for the computed time step if it is not being calculated? Does this nullptr check occur on processors with no nodes?

src/tribol/interface/tribol.cpp
srwopschall: numCells comes from the registered mesh. If there are no boundary faces on a particular processor then this should be zero.

src/tribol/interface/tribol.cpp
```diff
@@ -457,28 +460,25 @@ void registerNodalVelocities( integer meshId,
                  "no mesh with id, " << meshId << "exists.");
 
    MeshData & mesh = meshManager.GetMeshInstance( meshId );
+   mesh.m_nodalFields.m_is_velocity_set = true;
+
    if (vx == nullptr || vy == nullptr)
    {
-      mesh.m_isValid = false;
+      mesh.m_nodalFields.m_is_velocity_set = false;
```
liu15, line 467: should this be protected against numCells == 0?

src/tribol/mesh/CouplingScheme.cpp
srwopschall: The idea is that the 'dt' input argument is the actual dt used by the host-code for the current cycle. So in the case where Tribol doesn't actually compute a timestep (e.g. null velocity pointers), then we simply don't modify the existing dt. To your point, in case the host-code doesn't actually pass their current timestep, I could return a very large (i.e. infeasible) timestep value that a host-code may see, and decide not to use. Thoughts?

srwopschall: The nullptr check would occur on processors with zero-element meshes (e.g. no nodes). We simply wouldn't manipulate the dt input argument.

src/tribol/interface/tribol.cpp
srwopschall: I don't think we need to guard it here. When we check for a valid coupling scheme, we don't perform all initialization checks if numCells == 0 so this boolean is never checked. If numCells == 0, we only look to see if this boolean is true for the gap-rate method that NEEDS it. Otherwise, we 'tolerate' or allow null velocity pointers.

src/tribol/mesh/CouplingScheme.cpp
liu15: > 

If someone is using the timestep vote, they are probably just taking a minimum of the returned timestep, so they probably wouldn't notice.

I can think of two ways to handle this:
1) Have two different calls, one that returns the timestep vote and one that doesn't.  Only give the warning if the former is called.  Note that you need to make sure the warning is not emitted if the velocity is nullptr on a processor with no nodes.
2) Return a negative value for the timestep vote if the information provided is incomplete.  You would probably have to keep the invalid flag separate from the timestep estimate and do a separate reduction.  You would also have to make sure a processor with no nodes doesn't return invalid.

liu15: I'll let you think about that for a future TODO.

srwopschall: I think for now the single call is fine. I removed the warning message, so we tolerate null-velocity pointers. Null-pointers would arise from a user not registering them, in which case the dt input argument is left unchanged, and the other if the pointers are registered, but happen to be null. Also in that case the dt input argument is unchanged. Rather than not changing the dt, which should be set at the current host-code's cycle dt, we could return a negative value allowing the host-code to query if the timestep vote was successful. Alternatively, while we always attempt to compute a tribol timestep vote, we could have the user set a Tribol option to compute it or not. On-rank a host-code could opt out of computing it for ranks with zero contact faces. Thoughts?

liu15 (approved):

src/tribol/mesh/CouplingScheme.cpp
liu15: I like returning the negative timestep if failed (assuming that this doesn't get set for a null ptr from an empty processor).

PR #18, opened 2023-08-09
Prevent fortran from being enabled when ~fortran

chapman39:
If `fc` is specified in your `spack.yaml compilers:` or `compilers.yaml`, it will set ENABLE_FORTRAN-- even if you have `fortran` variant disabled. It would make sense to only enable fortran if the variant is set.


ebchin (approved):
Thanks for the fix @chapman39!

srwopschall (approved):
Thanks @chapman39

PR #22, opened 2023-08-25
Mfem mortar example and test fixes

ebchin:
A few fixes for the mortar example and test based on testing in Serac.

### DIFF
--- a/src/examples/mfem_mortar_lm_patch.cpp
+++ b/src/examples/mfem_mortar_lm_patch.cpp
@@ -394,7 +394,7 @@ int main( int argc, char** argv )
     R_submesh.Mult(g, G);
   }
 
-  // Create a single HypreParMatrix from blocks for the preconditioner
+  // Create a single HypreParMatrix from blocks (useful for different solvers/preconditioners)
   mfem::Array2D<mfem::HypreParMatrix*> hypre_blocks(2, 2);
   for (int i{0}; i < 2; ++i)
   {
--- a/src/examples/mfem_mortar_lm_patch.cpp
+++ b/src/examples/mfem_mortar_lm_patch.cpp
@@ -43,6 +43,9 @@
  * Example output can be viewed in VisIt or ParaView.
  */
 
+#include <mfem/linalg/hypre.hpp>
+#include <mfem/linalg/solvers.hpp>
+#include <mfem/linalg/superlu.hpp>
 #include <set>
 
 #ifdef TRIBOL_USE_UMPIRE
@@ -65,6 +68,12 @@
 #include "tribol/config.hpp"
 #include "tribol/interface/tribol.hpp"
 #include "tribol/interface/mfem_tribol.hpp"
+
+void clearInactivePressureDofs(
+  mfem::BlockOperator& A,
+  int coupling_scheme_id,
+  const std::set<int>& nonmortar_attrs
+);
 
 int main( int argc, char** argv )
 {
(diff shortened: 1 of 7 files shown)


src/examples/mfem_mortar_lm_patch.cpp
```diff
@@ -348,6 +357,9 @@ int main( int argc, char** argv )
   // HypreParMatrices holding the Jacobian from contact and stored in an MFEM
   // block operator are returned from this API call.
   auto A_blk = tribol::getMfemBlockJacobian(coupling_scheme_id);
+  // Surface mesh contains mortar and nonmortar surfaces.  Remove pressure DOFs
+  // on mortar surfaces.
+  clearInactivePressureDofs(*A_blk, coupling_scheme_id, nonmortar_attrs);
```
srwopschall: Does this clear just the mortar dofs? Or does this also clear inactive dofs that are non-mortar, but not part of the active set? Maybe clarify.

src/examples/mfem_mortar_lm_patch.cpp
```diff
@@ -369,23 +381,45 @@ int main( int argc, char** argv )
...
+  //
+  // NOTE: gap is a dual field, but we still apply R here because tribol stores
+  // this like a ParGridFunction: shared DOFs have the same (summed) value on
+  // all ranks
   {
     auto& G = B_blk.GetBlock(1);
     auto& R_submesh = *tribol::getMfemPressure(coupling_scheme_id)
-      .ParFESpace()->GetProlongationMatrix();
-    R_submesh.MultTranspose(g, G);
+      .ParFESpace()->GetRestrictionMatrix();
```
srwopschall, line 389: This looks like a bug fix?

src/examples/mfem_mortar_lm_patch.cpp
```diff
@@ -369,23 +381,45 @@ int main( int argc, char** argv )
...
+      {
+        hypre_blocks(i, j) = dynamic_cast<mfem::HypreParMatrix*>(&A_blk->GetBlock(i, j));
+      }
+      else
+      {
+        hypre_blocks(i, j) = nullptr;
+      }
+    }
+  }
+  auto A_merged = std::unique_ptr<mfem::HypreParMatrix>(
```
srwopschall, line 413: Since this is an example, maybe state what you mean by "merged".

src/tests/tribol_mfem_mortar_lm.cpp
```diff
@@ -255,3 +264,84 @@ int main(int argc, char* argv[])
 
   return result;
 }
+
+void clearInactivePressureDofs(
```
srwopschall: Sounds like you are going to move this to a central location for use by Tribol when using mortar methods with MFEM API. I don't think we need to expose this to the user, but it does sound like a needed utility function.

srwopschall (commented):
@ebchin - Looks pretty good. I made some minor comments. I didn't review the guts of the "clear" routine yet. I'll wait until you drop it somewhere central.

src/examples/mfem_mortar_lm_patch.cpp
ebchin: Sounds good.  It puts a 1 on the diagonal of mortar dofs in the matrix, but doesn't touch non-mortar inactive dofs.  As implemented, the host code is responsible for updating the active set.  Will update the comment on L359; the `clearInactivePressureDofs()` will be buried in `tribol::getMfemBlockJacobian()` after refactoring.

src/examples/mfem_mortar_lm_patch.cpp
ebchin: Correct, the added comment above explains why the fix is necessary

srwopschall (approved):

PR #23, opened 2023-08-27
Check to make sure DOFs exist on rank

ebchin:
Add a check in the Tribol MFEM interface to only try to get pointers to DOFs when the redecomp grid function has DOFs.

### DIFF
--- a/src/tribol/mesh/MfemData.cpp
+++ b/src/tribol/mesh/MfemData.cpp
@@ -699,7 +699,7 @@ void MfemJacobianData::UpdateJacobianXfer()
 }
 
 std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian(
-  const MethodData& method_data
+  const MethodData* method_data
 ) const
 {
   // 0 = displacement DOFs, 1 = lagrange multiplier DOFs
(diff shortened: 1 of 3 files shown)


src/tribol/mesh/MfemData.cpp
```diff
@@ -696,43 +708,53 @@ std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian(
   // (0,1) and (1,0) are symmetric (for now using SINGLE_MORTAR with approximate tangent)
   const auto& elem_map_1 = parent_data_.GetElemMap1();
   const auto& elem_map_2 = parent_data_.GetElemMap2();
-  auto mortar_elems = method_data
```
ebchin, line 699: This stuff is a bit of a hack because `MethodData` does not exist when the rank has no mesh.

It might make more sense to create `MethodData` even if the rank doesn't have a mesh.  It would get rid of the branching here and create a more unified code path for all ranks.  The downside is you'll have to allocate things unnecessarily (as noted [here](https://github.com/LLNL/Tribol/blob/develop/src/tribol/mesh/CouplingScheme.cpp#L975-L977)), but the rank where things are being allocated to is likely going to have to wait anyway for other ranks to finish processing their data.  Let's discuss.

srwopschall (approved):
@ebchin - Looks good. Add a comment as to why empty method_data stuff needs to be 'allocated' for mfem-based implicit contact methods.

PR #28, opened 2023-09-30
fix to always update m_interfacePairs

ebchin:
Fix to update m_interfacePairs when pair is not in contact


srwopschall (approved):
@ebchin - This looks good. Anymore modifications/refactor should probably go in a separate PR/issue as discussed.

PR #29, opened 2023-10-03
Feature/swopscha/logging

### LINKED ISSUE #14: Logging Revamp
- Update all logging to use SLIC, and use SLIC_<>_ROOT where appropriate so only rank = 0 outputs info, warnings, and errors.
- Add verbosity options? Suppress warnings?

srwopschall:
Updating logging with various SLIC levels:
- introduce SLIC_<>_ROOT where appropriate
- move various warnings to SLIC_INFO or SLIC_DEBUG
- ensure SLIC_WARNINGS are user appropriate
- modify error messages


src/tribol/geom/GeomUtilities.cpp
```diff
@@ -585,9 +585,13 @@ void Intersection2DPolygon( const real* const RESTRICT xA,
...
-   if (!orientA || !orientB)
+   if (!orientA)
    {
-     TRIBOL_ERROR("Intersection2DPolygon: polygon vertices not ordered counter-clockwise");
+      SLIC_ERROR("polygonal A vertices not CCW");
+   }
+   if (!orientB)
+   {
+      SLIC_ERROR("polygonal B vertices not CCW");
    }
```
kennyweiss: Gut check: Is this an ERROR, i.e. where we'd want the application to terminate for any pair of polygons w/ the wrong orientations? Seems like we could get into trouble with corner cases w/ bad areas/tolerances.

Alternatively, would we want to mark a flag as dirty and output a single warning message at the end of the loop that some pairs of polygons had bad orientations?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -332,7 +332,7 @@ void GlobalTo2DLocalCoords( const real* const RESTRICT pX,
 
    if (pLX == nullptr || pLY == nullptr)
    {
-      TRIBOL_ERROR ("GlobalTo2DLocalCoords: local coordinate pointers are null");
+      SLIC_ERROR ("GlobalTo2DLocalCoords: local coordinate pointers are null");
```
kennyweiss: Gut check: should this only be an error if `size > 0` ?
(and similarly elsewhere)

src/tribol/geom/ContactPlane.cpp
```diff
@@ -1129,7 +1101,7 @@ void ContactPlane3D::computeAreaTol()
    parameters_t & parameters = parameters_t::getInstance();
 
    if (m_areaFrac < parameters.overlap_area_frac ) {
-      SLIC_ERROR ( "ContactPlane3D::computeAreaTol the overlap area fraction too small or negative" );
+      SLIC_ERROR ( "ContactPlane3D::computeAreaTol() the overlap area fraction too small or negative" );
```
kennyweiss: Gut check -- is this an error that would require terminating the application? Or would it be better to log a debug message and `return 0;`?

src/tribol/geom/ContactPlane.cpp
```diff
@@ -1439,7 +1411,7 @@ bool ContactPlane3D::computeLocalInterpenOverlap()
 
          if (k > 2)
          {
-            SLIC_ERROR ("ContactPlane3D::computeInterpenOverlap : too many segment-plane intersections.");
+            SLIC_ERROR ("ContactPlane3D::computeInterpenOverlap(): too many segment-plane intersections.");
```
kennyweiss: Gut check -- this seems like an assertion (i.e. programmer check) but not necessarily a runtime error -- perhaps we'd want to just return early and/or handle this more gracefully in release builds.

src/tribol/geom/ContactPlane.cpp
```diff
@@ -2149,7 +2121,7 @@ void ContactPlane2D::computeAreaTol()
    // Note: this code is the same as for ContactPlane3D, but maintain separate 
    // routine for 2D treatment.
    if (m_areaFrac < 1.E-12) {
-      SLIC_ERROR ( "ContactPlane2D::computeAreaTol the overlap area fraction too small or negative" );
+      SLIC_ERROR ( "ContactPlane2D::computeAreaTol() the overlap area fraction too small or negative" );
```
kennyweiss: Again here -- is this a runtime error, or something we should handle more gracefully in release configs?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -818,7 +812,7 @@ void Intersection2DPolygon( const real* const RESTRICT xA,
          // debug
          if (k > (numPolyVert))
          {
-            TRIBOL_ERROR("Intersection2DPolygon: iterator k greater than numPolyVert (1)");
+            SLIC_DEBUG("Intersection2DPolygon: iterator k greater than numPolyVert (1)");
```
kennyweiss: Would more context be useful for debugging this?

kennyweiss: Also, `SLIC_DEBUG_IF( <cond>, <msg>)` might be more concise

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1137,10 +1126,9 @@ bool SegmentIntersection2D( const real xA1, const real yA1, const real xB1, cons
      yDiff = (yDiff < 0.) ? -1.0 * yDiff : yDiff;
 
      real diffTol = 1.0E-3;
-     SLIC_ERROR_IF( xDiff > diffTol || yDiff > diffTol, 
+     SLIC_DEBUG_IF( xDiff > diffTol || yDiff > diffTol, 
                    "SegmentIntersection2D(): Intersection coordinates are not equally derived." );
```
kennyweiss, line 1120: Future -- this message would likely be more useful w/ more context and/or actual values

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1275,10 +1259,9 @@ void CheckPolySegs( const real* const RESTRICT x, const real* const RESTRICT y,
    {
       if (newIDs[i] == i)
       {
-         // debug error
          if (k > numNewPoints)
          {
-            TRIBOL_ERROR("checkPolySegs: index into polyX/polyY exceeds allocated space");
+            SLIC_ERROR("checkPolySegs: index into polyX/polyY exceeds allocated space");
```
kennyweiss: Is there a way to handle this (more) gracefully?

src/tribol/interface/tribol.cpp
```diff
@@ -86,6 +86,11 @@ void initialize( integer dimension, CommType comm )
    parameters.problem_comm = comm;
 
    internal::set_defaults( );
+
+   // set default logging level
+   axom::slic::finalize(); // TODO do we need to finalize here
```
kennyweiss: Does this call only happen once? or every cycle?

I don't think we want to finalize or initialize slic in general unless the user asks us to since slic is a singleton that is shared with other libraries.

src/tribol/interface/tribol.cpp
```diff
@@ -96,9 +101,9 @@ void setPenaltyOptions( int couplingSchemeIndex, PenaltyConstraintType pen_enfrc
    CouplingSchemeManager& csManager = CouplingSchemeManager::getInstance();
 
    // check to see if coupling scheme exists
-   SLIC_ERROR_IF( !csManager.hasCoupling( couplingSchemeIndex ), 
-                  "tribol::setPenaltyOptions(): call tribol::registerCouplingScheme() " <<
-                  "prior to calling this routine." );
+   SLIC_ERROR_ROOT_IF( !csManager.hasCoupling( couplingSchemeIndex ),
```
kennyweiss, line 98: Note: The `_ROOT_` calls only work when a specific API call is invoked -- (`slic::setIsRoot()`)
See: https://github.com/LLNL/axom/blob/ca7e58ff86f5f16c4c1f57f3042bd911905adb13/src/axom/slic/interface/slic.hpp#L62-L68

kennyweiss, line 98: +1 using the ROOT macros is a good idea for messages that will by definition be the same on all ranks.

src/tribol/interface/tribol.cpp
```diff
@@ -1008,6 +1028,7 @@ void finalize( )
    {
       csManager.clearAllCouplings();
    }
+   axom::slic::finalize();
```
kennyweiss: I'm not sure that we want to call ``slic::finalize()`` here either. I think it should be the host-code's responsibility (i.e. at a higher level).

We might want to explicitly use ``slic::SimpleLogger`` in the tribol tests and examples, which will initialize and finalize slic as needed -- https://github.com/LLNL/axom/blob/develop/src/axom/slic/core/SimpleLogger.hpp

... or perhaps create a simple extension in tribol (or axom) that works with MPI

src/tribol/interface/tribol.hpp
```diff
@@ -144,6 +144,13 @@ void setPlotOptions( enum VisType v_type );
  */
 void setOutputDirectory( const std::string& dir );
 
+/*!
+ * \brief Sets the logging level
+ * \param [in] csId coupling scheme id
+ * \param [in] log_level the desired logging level 
+ */
+void setLoggingLevel( const int csId, const LoggingLevel log_level );
```
kennyweiss: Since Tribol has a hard dependency on axom and slic, any reason for it to define it's own `LoggingLevel` enum rather than use the slic one? If this planning for future extensions?

src/tribol/common/logger.hpp
```diff
@@ -7,15 +7,7 @@
...
 
 #ifdef USE_SLIC
-
 #include "axom/slic.hpp"  // for logging
-#define TRIBOL_ASSERT( x ) SLIC_ASSERT(x)
-#define TRIBOL_ERROR(x) SLIC_ERROR(x)
-#else
-
-#define TRIBOL_ASSERT( x )
-#define TRIBOL_ERROR(x)
```
kennyweiss, line 17: Thank you for removing this!

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -253,22 +253,22 @@ void CouplingSchemeInfo::printCaseInfo()
    {
       case SPECIFYING_NO_SLIDING_WITH_REGISTERED_MODE:
       {
-         SLIC_INFO("Overriding with ContactCase=NO_SLIDING with registered ContactMode."); 
+         SLIC_DEBUG("Overriding with ContactCase=NO_SLIDING with registered ContactMode.");
```
kennyweiss: Should these be `SLIC_DEBUG_ROOT`?

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -956,6 +962,8 @@ int CouplingScheme::apply( integer cycle, real t, real &dt )
 //------------------------------------------------------------------------------
 bool CouplingScheme::init()
 {
+   this->setSlicLoggingLevel();
```
kennyweiss: If we're changing the logging level, we should restore it at the end of the function.

Unfortunately, axom's `ScopedLogLevelChanger` has not yet been added to slic, 
(see: https://github.com/LLNL/axom/issues/1078)

kennyweiss: I think we should add something similar as a utility class until it gets into axom

See: https://github.com/LLNL/axom/blob/ca7e58ff86f5f16c4c1f57f3042bd911905adb13/src/axom/quest/interface/internal/QuestHelpers.hpp#L34-L61

Note: That class uses RAII, so it will undo itself when it goes out of scope, which might not be the right call here.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -410,14 +406,12 @@ int ApplyNormal< COMMON_PLANE, PENALTY >( CouplingScheme const * cs )
...
-           dbg_sum_force2 += magnitude( nodal_force_x2,
-                                        nodal_force_y2,
-                                        nodal_force_z2 );
-        #endif /* TRIBOL_DEBUG_LOG */
+        dbg_sum_force1 += magnitude( nodal_force_x1, 
+                                     nodal_force_y1, 
+                                     nodal_force_z1 );
+        dbg_sum_force2 += magnitude( nodal_force_x2,
+                                     nodal_force_y2,
+                                     nodal_force_z2 );
```
kennyweiss: Do you expect this to have a noticeable affect on performance? If so, it's probably be best to only compute it in debug configs as it was before.

kennyweiss (commented):
Thanks @srwopschall -- this looks like it's coming along nicely.

I sprinkled a few comments throughout.
The main things I'm concerned about are:
* initializing and finalizing slic, which might be setup differently by a calling code
* modifying the logging level locally without restoring it

kennyweiss:
Tag: @bmhan12  -- `tribol` uses `slic` pretty extensively

src/tribol/geom/GeomUtilities.cpp
srwopschall: The logging in the computational geometry routines has been refactored to push 'single messages' to higher level calling routines. This still could be refined, but this reduces lots of duplicate or extraneous logging. Moreover, many of the messages, such as the ones pointed out here, have been reduced to `SLIC_DEBUG` calls. Additionally, this orientation check has been refactored. Somewhat confusing, but sometimes you want to check and log a message if appropriate, and other times you do not.

src/tribol/geom/GeomUtilities.cpp
srwopschall: Good catch. Fixed.

src/tribol/geom/ContactPlane.cpp
srwopschall: I fixed this to report a debug message.  I fixed up the checks for the area fraction too.

src/tribol/geom/ContactPlane.cpp
srwopschall: This sort of thing can likely be refined, but I refactored this to use `SLIC_INFO` and return early with an error enum.

src/tribol/geom/ContactPlane.cpp
srwopschall: I modified this to be a debug message and have modified how the area fraction is set and checked.

src/tribol/geom/GeomUtilities.cpp
srwopschall: Fixed.

src/tribol/geom/GeomUtilities.cpp
srwopschall: True. I'll hold off for now. This is really more of a math sanity check and may not necessarily need to be checked. It has never presented an issue.

src/tribol/geom/GeomUtilities.cpp
srwopschall: I fixed this, and another location where this iterator 'may' exceed allocated space (i.e. number of polygonal overlap vertices). I added an enum to catch and report this at a higher level.

src/tribol/interface/tribol.cpp
srwopschall: Correct. I removed the call to initialize and finalize. In general, these calls will be handled by the host-code. I am not going to introduce the ability for a user to initialize or finalize slic for now since this shouldn't be the general use case.

src/tribol/interface/tribol.cpp
srwopschall: I added this to `tribol::initialize()` (I believe correctly).

src/tribol/interface/tribol.cpp
srwopschall: I removed this. And `slic::SimpleLogger` is currently being used in tribol tests and examples.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Good catch. Fixed.

src/tribol/interface/tribol.hpp
srwopschall: I was staying flexible by doing this, but I don't see any need for a future extension. Modify to use the slic namespace logging levels?

src/tribol/mesh/CouplingScheme.cpp
srwopschall: This has been completely refactored and is controlled at the tribol API level. You set a logging level per coupling scheme. This seems most natural, but also flexible in that one could in principle debug a particular coupling scheme? Alternatively, we can set one logging level for ALL coupling schemes. Thoughts?

src/tribol/physics/CommonPlane.cpp
srwopschall: I removed the `TRIBOL_DEBUG_LOG` which guarded these calculations. It isn't perfect, but I added an if statement around these querying the coupling scheme logging level and only compute these if the logging level is DEBUG. Can refactor this with ifdef later?

srwopschall:
@kennyweiss , @ebchin , @rieben1 , @white238 - This PR is ready for review. The main thrust is to reduce the amount of logging various physics packages may experience during testing. This utilizes SLIC macros and reduced a lot of warnings to lesser levels. We can set a single logging level through `initialize()` or optionally we have the ability to set the logging level per coupling scheme through a new API function. Ultimately it will be the host-codes responsibility to reset the logging level if the tribol one differs.

src/tests/tribol_mortar_data_geom.cpp
```diff
@@ -137,7 +137,7 @@ TEST_F( MortarGeomTest, mortar_good_patch )
    i_ys.close();
    i_zs.close();
 
-   SLIC_INFO("After loading mesh data and constructing mfem vectors.");
+   SLIC_DEBUG("After loading mesh data and constructing mfem vectors.");
```
kennyweiss: Minor: These actually seem like reasonable `SLIC_INFO` statements to me -- they're only called once and they let you know where the application is. 

Making them `SLIC_DEBUG` is also reasonable, so, no need to change them back if you disagree.

src/tribol/geom/ContactPlane.hpp
```diff
@@ -30,15 +30,17 @@ class ContactPlaneManager;
  * \param [in] pair interface pair containing pair related indices
  * \param [in] cMethod the Tribol contact method
  * \param [in] cCase the Tribol contact Case
+ * \param [in] inContact true if pair are in contact per CG routines
  *
- * \return true if face-pair are interacting
+ * \return 0 if no error, 1 if error
```
kennyweiss: (Here and elsewhere) it might be more accurate/helpful to say something like
```
return 0 if no error; non-zero (via FaceGeomError enum) otherwise
```

kennyweiss: Note: This is a breaking change that should probably be mentioned in the `RELEASE-NOTES`. I think it's mostly an internal function call, so it probably won't impact that much.

src/tribol/interface/simple_tribol.cpp
```diff
@@ -31,20 +31,11 @@ namespace numerics = axom::numerics;
...
-      axom::slic::initialize();
-      std::string format = "[<LEVEL>]: <MESSAGE> \n";
-      axom::slic::setLoggingMsgLevel( axom::slic::message::Info );
-
-      axom::slic::addStreamToAllMsgLevels(
-         new axom::slic::GenericOutputStream( &std::cout,format) );
-   }
-
+   // Note, slic initialization moved into tribol::initialize(). 
+   // Function prototype not modified due to shroud layer
```
kennyweiss: 👍

kennyweiss: If we no longer need the `init_slic` parameter, we can (and should) change the shroud interface in a follow-up PR.

src/tribol/interface/tribol.hpp
kennyweiss: I think that makes sense -- since this PR is time-sensitive perhaps we can make that change in a follow-up PR.

src/tribol/utils/TestUtils.cpp
```diff
@@ -1174,6 +1174,10 @@ int TestMesh::tribolSetupAndUpdate( ContactMethod method,
                            enforcement,
                            BINNING_GRID );
 
+   std::cout << "Before tribol::setLoggingLevel()." << std::endl;
+   setLoggingLevel(csIndex, WARNING);
+   std::cout << "After tribol::setLoggingLevel()." << std::endl;
```
kennyweiss: Should we remove these `cout`s before merging? I assume they're only here for local debugging.

src/tribol/geom/ContactPlane.cpp
```diff
@@ -46,29 +48,55 @@ bool CheckInterfacePair( InterfacePair& pair,
          // get instance of contact plane manager to store all contact plane data per cycle
          ContactPlaneManager& cpMgr = ContactPlaneManager::getInstance();
 
+         // set whether full overlap is to be used or not. Note: SINGLE_MORTAR and 
+         // MORTAR_WEIGHTS drop into this 'case', so the method still needs to be checked
          bool full = (cMethod == COMMON_PLANE) ? false : true;
```
kennyweiss: Minor: can some/all of the vars in this block be `const`?

src/tribol/geom/ContactPlane.cpp
```diff
@@ -1439,7 +1440,9 @@ bool ContactPlane3D::computeLocalInterpenOverlap()
 
          if (k > 2)
          {
-            SLIC_ERROR ("ContactPlane3D::computeInterpenOverlap : too many segment-plane intersections.");
+            SLIC_INFO("ContactPlane3D::computeInterpenOverlap(): too many segment-plane intersections.");
```
kennyweiss: Gut-check: Should this be a `SLIC_INFO`? Would `SLIC_DEBUG` work?

Now that it's no longer marked as an error and the return type is `DEGENERATE_OVERLAP`, I'm wondering if this might lead to too many log messages in some cases

kennyweiss: Also, could/should we add more context about the specific error? 
There's not too much to go on here if we do encounter this problem.

src/tribol/geom/ContactPlane.cpp
```diff
@@ -2148,8 +2157,11 @@ void ContactPlane2D::computeAreaTol()
 {
    // Note: this code is the same as for ContactPlane3D, but maintain separate 
    // routine for 2D treatment.
-   if (m_areaFrac < 1.E-12) {
-      SLIC_ERROR ( "ContactPlane2D::computeAreaTol the overlap area fraction too small or negative" );
+   parameters_t & parameters = parameters_t::getInstance();
+
+   if (m_areaFrac < parameters.overlap_area_frac ) {
+      SLIC_DEBUG( "ContactPlane2D::computeAreaTol() the overlap area fraction too small or negative; " << 
+                  "setting to overlap_area_frac parameter." );
```
kennyweiss: Minor: Can/should we use `SLIC_DEBUG_IF` here instead?

src/tribol/geom/ContactPlane.cpp
```diff
@@ -2275,7 +2287,9 @@ bool ContactPlane2D::computeLocalInterpenOverlap()
    // Debug check the number of interpenetrating vertices
    if (k > 2)
    {
-      SLIC_ERROR("ContactPlane2D::computeLocalInterpenOverlap more than 2 interpenetrating vertices detected.");
+      SLIC_INFO("ContactPlane2D::computeLocalInterpenOverlap() more than 2 interpenetrating vertices detected.");
```
kennyweiss: Same question here -- should this be a `SLIC_INFO` and, if so, should it have more context about the problem?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -330,9 +330,9 @@ void GlobalTo2DLocalCoords( const real* const RESTRICT pX,
                             real* const RESTRICT pLY, int size )
 {
 
-   if (pLX == nullptr || pLY == nullptr)
+   if (size > 0 && (pLX == nullptr || pLY == nullptr))
```
kennyweiss: Minor (clarity): a few of these can be `SLIC_ERROR_IF` since there's no other code inside the curly braces

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -760,10 +759,10 @@ void Intersection2DPolygon( const real* const RESTRICT xA,
          // if both segments are not defined by nodes interior to the other polygon
          if (checkA && checkB) 
          {
-            // debug
             if (interId >= maxSegInter) 
             {
-               SLIC_ERROR ("Intersection2DPolygon: number of segment intersections exceeds precomputed maximum.");
+               SLIC_INFO("Intersection2DPolygon: number of segment intersections exceeds precomputed maximum.");
```
kennyweiss: Gut-check: Does this need to be a `SLIC_INFO` and, if so, could the message be improved?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1474,6 +1470,26 @@ void PolyReorder( real* const RESTRICT x, real* const RESTRICT y, const int numP
 
 } // end PolyReorder()
 
+//------------------------------------------------------------------------------
+void PolyReverse( real* const RESTRICT x, real* const RESTRICT y, const int numPoints )
+{
+   real xtemp[ numPoints ];
```
kennyweiss, line 1448: Warning: Runtime-sized arrays are non-standard and might fail w/ certain compilers and/or flags.

This operation can be done in place using something like this: 
https://github.com/LLNL/axom/blob/c35431a1ee71a7c46fe54805e8aa5278120d613e/src/axom/primal/geometry/BezierCurve.hpp#L296-L301

src/tribol/interface/tribol.cpp
```diff
@@ -608,13 +671,13 @@ int getJacobianCSRMatrix( int** I, int** J, real** vals, int csId,
       }
       case SINGLE_MORTAR:
       {
-         SLIC_WARNING("tribol::getCSRMatrix(): CSR format not currently implemented with "
+         SLIC_WARNING("tribol::getJacobianCSRMatrix(): CSR format not currently implemented with "
```
kennyweiss, line 636: Should these be `SLIC_WARNING_ROOT` to avoid excessive messages?

src/tribol/interface/tribol.cpp
```diff
@@ -961,17 +1024,19 @@ integer update( integer cycle, real t, real &dt )
...
          if (couplingScheme->nullMeshes())
          {
             continue;
          }
          else
          {
-            SLIC_WARNING("Skipping invalid CouplingScheme " << couplingScheme->getId() << ". " << 
-                         "Please see warnings.");
+            SLIC_INFO("tribol::update(): skipping invalid CouplingScheme " << 
+                      couplingScheme->getId() << "Please see warnings.");
```
kennyweiss: Please explain why this is no longer a warning -- is it b/c it's expected behavior for some user codes?

kennyweiss (approved):
Thanks @srwopschall -- I had several minor comments. One potential problem is related to the variable-sized array in the reverse orientation function, since that's non-standard.

src/tests/tribol_mortar_data_weights.cpp
```diff
@@ -93,19 +90,19 @@ void TestMortarWeights( tribol::CouplingScheme const * cs, double exact_area, do
           
          if ( J[b] < nodeOffset ) // nonmortar/mortar  weight
```
ebchin: Is there a `SLIC_DEBUG_IF` to get rid of the if statements?

src/tests/tribol_mortar_gap.cpp
```diff
@@ -163,6 +163,9 @@ class MortarGapTest : public ::testing::Test
          }
       }
 
+      tribol::CommType problem_comm = TRIBOL_COMM_WORLD;
+      tribol::initialize( 3, problem_comm );
```
ebchin, line 167: Does `finalize()` need to be called somewhere and/or should the call to `initialize()` be in `main()`?

src/tests/tribol_tet_mesh.cpp
```diff
@@ -148,7 +148,7 @@ TEST_F( TetMeshTest, build_and_check_mfem_tet_mesh )
```
ebchin, line 148: Does this source file need a `SimpleMPIWrapper`?  (GitHub won't let me put this comment further down where it belongs)

src/tribol/common/logger.hpp
```diff
@@ -7,15 +7,9 @@
...
-
 #endif
 
+// TODO keep file for now to allow finer grained logging in the future
```
ebchin: I say we remove this file for now to keep things clean.  We will likely need some design discussions before we implement more fine grained logging which may or may not include this file.

src/tribol/geom/ContactPlane.hpp
```diff
@@ -30,15 +30,17 @@ class ContactPlaneManager;
  * \param [in] pair interface pair containing pair related indices
```
ebchin, line 30: I think this routine modifies this?  Should it be `param [in,out]`?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -30,15 +30,17 @@ class ContactPlaneManager;
...
  * \param [in] cMethod the Tribol contact method
  * \param [in] cCase the Tribol contact Case
+ * \param [in] inContact true if pair are in contact per CG routines
```
ebchin, line 33: Is this just a param in?  If so, it should probably be `const`.

src/tribol/geom/ContactPlane.hpp
ebchin: When we take care of #31, we might want to 
1) think about how each function interacts with the structs.  It would be good for each function to return its own struct rather than modify an existing one if possible, so the state of the struct is clear.
2) add some more documentation to methods like this describing what checks it is doing.

src/tribol/geom/ContactPlane.cpp
```diff
@@ -603,22 +614,13 @@ ContactPlane3D CheckFacePair( InterfacePair& pair,
...
-      {
-         X2[k] = projeX2[i];
-         Y2[k] = projeY2[i];
-         ++k;
-      }
+      real* X2 = &projeX2[0];
+      real* Y2 = &projeY2[0];
+ 
+      // assuming each face's vertices are ordered WRT that face's outward unit normal,
+      // reorder face 2 vertices to be consistent with face 1. DO NOT CALL POLYREORDER()
```
ebchin, line 621: This comment is confusing to me.  Maybe just say reverse the order of face 2's vertices so they're CCW w.r.t. the orientation of face 1?  Maybe this can be moved into `Intersection2DPolygon()`?  Also, does this need to be done?

src/tribol/geom/ContactPlane.cpp
```diff
@@ -542,7 +558,6 @@ ContactPlane3D CheckFacePair( InterfacePair& pair,
    real projX1[ mesh1.m_numNodesPerCell ];
```
ebchin, line 558: I think the above comment needs to be updated.  In Puso and Laursen, the contact plane is defined by the average normal of the nonmortar face.

src/tribol/common/Parameters.hpp
```diff
@@ -275,6 +288,19 @@ enum BasisEvalType
    NUM_BASIS_EVAL_TYPES = PHYSICAL
 };
 
+/*!
+ * \brief Enumerates face-pair computational geometry errors
+ */
+enum FaceGeomError
```
ebchin, line 294: Can this be a scoped enum (an `enum class`)?  It would help with readability in other source files that use this enum.

src/tribol/geom/ContactPlane.hpp
```diff
@@ -220,8 +222,12 @@ class ContactPlane
 
    /*!
     * \brief Compute the contact plane integral gap expression
+    *
+    * \param [in] interpen true if the two faces interpenetrate
```
ebchin: `[in,out]`?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -558,8 +564,12 @@ class ContactPlane3D : public ContactPlane
     * \brief Computes the polygonal overlap between the portion of two  
     *        faces lying on the contact plane that are interpenetrating the contact 
     *        plane.
+    *
+    * \param [in] interpen true if the two faces interpenetrate
```
ebchin: `[in,out]`?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -659,8 +669,12 @@ class ContactPlane2D : public ContactPlane
...
```
ebchin: `[in,out]`?

src/tribol/geom/ContactPlane.cpp
ebchin: I think `SLIC_DEBUG` makes sense here.  I imagine this might happen when the projection is non-convex, which could happen pretty frequently if the faces are nearly orthogonal.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -890,22 +894,34 @@ int CouplingScheme::apply( integer cycle, real t, real &dt )
...
 
-     if (!interact)
+
+     // TODO refine how these errors are handled. Here we skip over face-paris with errors. That is, 
+     // they are not registered for contact, but we don't error out.
+     if (interact_err != NO_FACE_GEOM_ERROR)
+     {
+        pair_err = 1;
+        pair.inContact = false;
+        SLIC_DEBUG("Face geometry error, " << static_cast<int>(interact_err) << "for pair, " << kp << ".");
```
ebchin, line 921: Not necessarily for this PR, but let's add an option to print out the coordinates of the error-raising face pair (either to screen or to file).  That way we can 
1) interrogate them further
2) add them to tests
3) gain a better understanding of the geometries that trigger errors and handle them more elegantly and/or give more helpful error messages

src/tribol/geom/ContactPlane.cpp
```diff
@@ -1592,24 +1599,26 @@ bool ContactPlane3D::computeLocalInterpenOverlap()
                           m_cX, m_cY, m_cZ,
                           cfx2_loc, cfy2_loc, numV[1] );
 
-   // reorder polygon local vertices in CCW orientation
+   // reorder potentially unordered set of vertices
```
ebchin, line 1605: Thanks for clarifying what this does

src/tribol/geom/ContactPlane.cpp
ebchin: I think we also need to set `m_areaFrac = parameters.overlap_area_frac` in this if statement (based on what the `SLIC_DEBUG` comment says)?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -688,15 +702,15 @@ class ContactPlane2D : public ContactPlane
  * \brief Checks if face-pair (3D) candidate is actual local contact interaction.
  *
  * \param [in] pair interface pair containing pair related indices
```
ebchin, line 704: `[in,out]`?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -712,15 +726,15 @@ ContactPlane3D CheckAlignedFacePair( InterfacePair& pair );
  * \brief Checks if 2D edge-pair candidate is actual local contact interaction.
  *
  * \param [in] pair interface pair containing pair related indices
```
ebchin, line 728: `[in,out]`?

src/tribol/geom/ContactPlaneManager.cpp
```diff
@@ -743,7 +743,8 @@ void ContactPlaneManager::getContactPlaneOverlapVerts( int const id,
 {
```
ebchin, line 743: Can't add the comment where it belongs, but should the `SLIC_ASSERT` on L574 be a `SLIC_ERROR`?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1137,10 +1138,9 @@ bool SegmentIntersection2D( const real xA1, const real yA1, const real xB1, cons
      yDiff = (yDiff < 0.) ? -1.0 * yDiff : yDiff;
 
      real diffTol = 1.0E-3;
-     SLIC_ERROR_IF( xDiff > diffTol || yDiff > diffTol, 
+     SLIC_DEBUG_IF( xDiff > diffTol || yDiff > diffTol,
```
ebchin, line 1119: Maybe put the computations in the brackets in a function that returns a bool so it's only computed if `SLIC_DEBUG` messages are on?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1297,7 +1293,7 @@ void PolyReorder( real* const RESTRICT x, real* const RESTRICT y, const int numP
 
    if (numPoints < 3)
    {
-      TRIBOL_ERROR ("PolyReorder: numPoints < 3.");
+      SLIC_ERROR ("PolyReorder: numPoints < 3.");
```
ebchin: This one looks like a good candidate for `SLIC_ERROR_IF`

src/tribol/interface/tribol.cpp
```diff
@@ -86,6 +86,50 @@ void initialize( integer dimension, CommType comm )
...
+         axom::slic::setLoggingMsgLevel( axom::slic::message::Error );
+         break;
+      } 
+      default:
+      {
+         axom::slic::setLoggingMsgLevel( axom::slic::message::Warning );
+         break;
+      }
+   } // end switch
+   //axom::slic::addStreamToAllMsgLevels(
```
ebchin: Can these commented out lines be removed?

src/tribol/interface/tribol.cpp
```diff
@@ -86,6 +86,50 @@ void initialize( integer dimension, CommType comm )
...
+   // set default logging level
+//   std::string format = "[<LEVEL>]: <MESSAGE> \n";
+
+   int myRank = 0;
+   #ifdef TRIBOL_USE_MPI
+   MPI_Comm_rank(comm, &myRank);
+   #endif
+   if (myRank == 0)
+   {
+      axom::slic::setIsRoot(true);
```
ebchin: Should we let the host code handle this or make it optional?  They might have called `setIsRoot()` with some other rank and changing it in `initialize()` might lead to a bug in the host code that is hard to find.

src/tribol/interface/tribol.cpp
```diff
@@ -313,14 +363,30 @@ void setOutputDirectory( const std::string& dir)
...
 }
 
+//------------------------------------------------------------------------------
+void setLoggingLevel( const int csId, const LoggingLevel log_level )
+{
+   CouplingSchemeManager& csManager = CouplingSchemeManager::getInstance();
+   CouplingScheme* couplingScheme  = csManager.getCoupling( csId );
+   if ( !in_range(log_level, NUM_LOGGING_LEVELS) )
+   {
+      SLIC_INFO_ROOT("tribol::setLoggingLevel(): Logging level not an option; " <<
```
ebchin, line 339: Maybe we should just leave the logging level as default for this case?

src/tribol/interface/tribol.cpp
```diff
@@ -542,12 +611,8 @@ void registerNodalResponse( integer meshId,
 int getJacobianSparseMatrix( mfem::SparseMatrix ** sMat, int csId )
 {
 
-   if (*sMat != nullptr)
-   {
-      SLIC_WARNING("tribol::getMfemSparseMatrix(): sparse matrix pointer not null; " <<
-                   "nullifying now.");
-      *sMat = nullptr;
-   }
+   SLIC_ERROR_IF(*sMat!=nullptr, "tribol::getMfemSparseMatrix(): " <<
```
ebchin, line 574: Maybe a note why `SLIC_ERROR_IF` is used here instead of `SLIC_ERROR_ROOT_IF`?  I think it's because we're accessing `MethodData`, which may not exist on ranks with null meshes?

src/tribol/interface/tribol.cpp
```diff
@@ -339,18 +405,20 @@ void registerMesh( integer meshId,
...
    const int dim = (z == nullptr) ? 2 : 3;
 
-   // check for null pointers for non-null meshes
+   // check for null pointers on ranks with non-null meshes
    if (numCells > 0)
    {
       if (x == nullptr || y == nullptr)
       {
-         SLIC_WARNING("Pointer to x or y-component mesh coordinate arrays are null pointers " <<
+         SLIC_WARNING("tribol::registerMesh(): pointer to x or y-component " <<
```
ebchin: Maybe add the rank where this warning is happening?  Though if you leave the rank off, `SLIC_WARNING_ROOT` is probably OK here.

src/tribol/interface/tribol.cpp
```diff
@@ -656,14 +719,14 @@ void registerMortarGaps( integer meshId,
...
-                 "no mesh with id " << meshId << " exists.");
+   SLIC_ERROR_ROOT_IF(!meshManager.hasMesh(meshId), "tribol::registerMortarGaps(): " << 
+                      "no mesh with id " << meshId << " exists.");
 
    MeshData & mesh = meshManager.GetMeshInstance( meshId );
 
    if (gaps == nullptr)
    {
-      SLIC_WARNING( "tribol::registerMortarGaps(): null pointer to data " << 
+      SLIC_WARNING( "tribol::registerMortarGaps(): null pointer to gap data " <<
```
ebchin, line 691: Maybe `SLIC_DEBUG_ROOT` or getting rid of this warning altogether?  This will happen a lot in parallel.

src/tribol/interface/tribol.cpp
```diff
@@ -681,14 +744,14 @@ void registerMortarPressures( integer meshId,
...
-                 "no mesh with id " << meshId << " exists.");
+   SLIC_ERROR_ROOT_IF(!meshManager.hasMesh(meshId), "tribol::registerMortarPressures(): " << 
+                      "no mesh with id " << meshId << " exists.");
 
    MeshData & mesh = meshManager.GetMeshInstance( meshId );
 
    if (pressures == nullptr)
    {
-      SLIC_WARNING( "tribol::registerMortarPressures(): null pointer to data " << 
+      SLIC_WARNING( "tribol::registerMortarPressures(): null pointer to pressure data " <<
```
ebchin, line 716: Same as comment on gaps.

src/tribol/interface/tribol.cpp
```diff
@@ -217,6 +261,12 @@ void setContactPenFrac( double frac )
 void setContactAreaFrac( double frac )
 {
    parameters_t & parameters = parameters_t::getInstance();
+   if (frac < 1.e-12)
+   {
+      SLIC_DEBUG("tribol::setContactAreaFrac(): area fraction too small or negative; " <<
```
ebchin: `SLIC_DEBUG_ROOT`?

srwopschall:
Note: I removed the logging level input in tribol::initialize() in favor of letting the host code set this. This does not address the issue of a host-code not natively using slic, which can/will be addressed in a future PR. A host-code can still set the logging level per individual coupling scheme, but for now the assumption is that slic has been initialized by the host-code and will be finalized or handle resetting the logging level.

src/tribol/interface/tribol.cpp
```diff
@@ -313,14 +363,30 @@ void setOutputDirectory( const std::string& dir)
    // Create path if it doesn't already exist
    if(! axom::utilities::filesystem::pathExists(dir) )
```
ebchin, line 319: I assume so, but does `axom::utilities::filesystem` work in parallel?

src/tribol/interface/tribol.cpp
```diff
@@ -726,8 +789,8 @@ void registerRealElementField( integer meshId,
 {
    MeshManager & meshManager = MeshManager::getInstance();
 
-   SLIC_ERROR_IF(!meshManager.hasMesh(meshId), "tribol::registerRealElementField(): " << 
-               "no mesh with id " << meshId << " exists.");
+   SLIC_ERROR_ROOT_IF(!meshManager.hasMesh(meshId), "tribol::registerRealElementField(): " << 
+                      "no mesh with id " << meshId << " exists.");
 
    MeshData & mesh = meshManager.GetMeshInstance( meshId );
```
ebchin, line 758: I think all the `SLIC_WARNING`s in this function should be `SLIC_WARNING_ROOT`s

src/tribol/interface/tribol.hpp
srwopschall: We will keep the new enum for now per design discussion.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: I have modified things per our design discussion. We don't modify the logging for all coupling schemes (and tribol calls) in tribol::initialize() and let the host code control this. We 'can' modify logging per coupling scheme, but this will be enhanced in a future PR. The assumption is that the host-code handles the restoration of the logging level after tribol::update() and they handle all initialization and finalization. Enhancing this is out of the scope of this PR.

src/tribol/interface/tribol.cpp
```diff
@@ -837,8 +900,8 @@ void registerRealElementField( integer meshId,
       }
       default:
       {
-         SLIC_WARNING( "tribol::registerRealElementField(): the field argument " << 
-                       "on mesh " << meshId << " is not an accepted tribol real element field." );
+         SLIC_ERROR( "tribol::registerRealElementField(): the field argument " <<
```
ebchin, line 865: Probably should be `SLIC_ERROR_ROOT`. Also, might want to get rid of `UNDEFINED_REAL_ELEMENT_FIELD` and `NUM_REAL_ELEMENT_FIELDS` in the enum so this error isn't needed.

src/tests/tribol_mortar_data_geom.cpp
srwopschall: I changed them back to info. For this test I think that makes a little more sense.

src/tribol/interface/tribol.cpp
```diff
@@ -851,14 +914,14 @@ void registerIntElementField( integer meshId,
 {
```
ebchin, line 876: We can probably remove this method and the related enum for code clarity.  They can be reintroduced if needed.

src/tribol/interface/tribol.hpp
```diff
@@ -30,7 +30,7 @@ namespace tribol
  * \pre dimension==2 || dimensions==3
  * \pre comm != MPI_COMM_NULL
  */
-void initialize( integer dimension, CommType comm );
```
ebchin: Please update the doxygen comment

src/tribol/interface/tribol.hpp
```diff
@@ -144,6 +144,15 @@ void setPlotOptions( enum VisType v_type );
...
 void setOutputDirectory( const std::string& dir );
 
+/*!
+ * \brief Optionally sets the logging level per coupling scheme
+ * \param [in] csId coupling scheme id
+ * \param [in] log_level the desired logging level 
+ *
+ * \note this overrides the logging level set in initialize().
+ */
+void setLoggingLevel( const int csId, const LoggingLevel log_level );
```
ebchin: Since these aren't references, I think `const` can be removed from the function parameters.

src/tribol/geom/ContactPlane.hpp
srwopschall: Fixed.

src/tribol/interface/simple_tribol.cpp
srwopschall: I removed these comments because we actually don't have slic initialization in tribol::initialize() per design discussion.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -389,24 +386,24 @@ bool CouplingScheme::isValidCouplingScheme()
...
-      SLIC_WARNING("Coupling scheme, " << this->m_id << ", does not have valid meshes.");
       return false;
    }
    
-   // return early for coupling schemes with one or both null meshes. These are no-op coupling schemes
+   // set boolean for null meshes
    if ( mesh1.m_numCells <= 0 || mesh2.m_numCells <= 0 )
    {
-      SLIC_INFO("Coupling scheme, " << this->m_id << ", has null-mesh(es).");
+      SLIC_DEBUG("Coupling scheme, " << this->m_id << ", has null-mesh(es).");
```
ebchin: I'd suggest either removing this or printing the rank if you keep it.

src/tribol/interface/tribol.cpp
srwopschall: I removed this. We agreed that for now the host-code is responsible for all initialization, calls to setIsRoot() and finalize.

src/tribol/geom/ContactPlane.hpp
srwopschall: Updated.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -921,9 +937,17 @@ int CouplingScheme::apply( integer cycle, real t, real &dt )
 
    } // end loop over pairs
 
+   // TODO refine how this logging is handled. This just detects an issue with a face-pair geometry
+   // (which has been skipped over for contact eligibility) and reports this warning. Do we want to 
+   // error out, or let a user detect bad contact behavior, but with a contact interaction that still
+   // runs?
```
ebchin, line 943: If you ask me, we should report this back to the host code somehow and let them decide how to proceed.  We definitely should not error out.  In an implicit code doing Newton iterations, we'd likely want to try again with a smaller step.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -962,6 +997,8 @@ bool CouplingScheme::init()
    valid = this->isValidCouplingScheme();
    if (valid)
    {
+      // set individual coupling scheme logging level
+      this->setSlicLoggingLevel();
```
ebchin, line 993: It wasn't clear to me why this would be called in an `init()` method until I saw where `init()` was being called.  Maybe renaming the method to something like `setupBeforeApply()` might be clearer?

src/tribol/geom/GeomUtilities.cpp
srwopschall: Yes, good point. Fixed in this file.

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -333,6 +333,12 @@ class CouplingScheme
                              const integer cycle, 
                              const real t );
 
+  void setLoggingLevel( const LoggingLevel log_level ) { m_loggingLevel = log_level; }
```
ebchin, line 342: Please add doxygen comments for these methods

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -521,6 +527,8 @@ class CouplingScheme
   EnforcementMethod m_enforcementMethod; ///< Contact enforcement method
   BinningMethod m_binningMethod;         ///< Contact binning method
 
+  LoggingLevel m_loggingLevel;
```
ebchin: Add doxygen comment

src/tribol/geom/GeomUtilities.cpp
srwopschall: I changed to `SLIC_DEBUG`. I think the message is useful, but not at the INFO level. I also improved the message.

src/tribol/geom/GeomUtilities.cpp
srwopschall: I will make an issue. This is outside the scope of this PR, but does need to be addressed.

src/tribol/physics/Mortar.cpp
```diff
@@ -95,12 +95,10 @@ void ComputeMortarWeights( SurfaceContactElem & elem )
             LinIsoQuadShapeFunc( xi[0], xi[1], a, phiNonmortarA );
             LinIsoQuadShapeFunc( xi[0], xi[1], b, phiNonmortarB );
 
-            #ifdef TRIBOL_DEBUG_LOG
-               if (nonmortarNonmortarId > elem.numWts || mortarNonmortarId > elem.numWts)
-               {
-                  TRIBOL_ERROR("ComputeMortarWts: integer ids for weights exceed elem.numWts");
-               }
-            #endif /* TRIBOL_DEBUG_LOG */
+            if (nonmortarNonmortarId > elem.numWts || mortarNonmortarId > elem.numWts)
```
ebchin: `SLIC_ERROR_IF`?

src/tribol/physics/Mortar.cpp
```diff
@@ -123,12 +121,10 @@ void ComputeNodalGap< SINGLE_MORTAR >( SurfaceContactElem & elem )
 {
    // check to make sure mortar weights have been computed locally 
    // for the SurfaceContactElem object
-   #ifdef TRIBOL_DEBUG_LOG
-      if (elem.mortarWts == nullptr)
-      {
-         TRIBOL_ERROR("ComputeNodalGap< SINGLE_MORTAR >: compute local weights on input struct first.");
-      }
-   #endif /* TRIBOL_DEBUG_LOG */
+   if (elem.mortarWts == nullptr)
```
ebchin: `SLIC_ERROR_IF`?

ebchin: There are a few more I see: L137, 142

src/tribol/interface/tribol.cpp
srwopschall: I'm going to leave this for now. I don't know if ranks with null-meshes will always call this routine. Sort of depends on how the host-code implements it. I think more implicit integration/testing will shed light on this.

src/tribol/utils/ContactPlaneOutput.cpp
```diff
@@ -98,6 +98,7 @@ void WriteContactPlaneMeshToVtk( const std::string& dir, const VisType v_type,
             overlaps = true;
```
ebchin, line 98: I think the default rank should be 0 (see L65)?

src/tribol/utils/TestUtils.cpp
```diff
@@ -773,9 +773,9 @@ void TestMesh::allocateAndSetVelocities( int meshId, real valX, real valY, real
    // Check that mesh ids are not the same. The TestMesh class was built around 
    // testing the mortar method with Lagrange multiplier enforcement, which does not 
    // support auto contact.
-   SLIC_WARNING_IF( this->mortarMeshId == this->nonmortarMeshId, 
-                    "TestMesh::allocateAndSetVelocities(): please set unique " << 
-                    "mortarMeshId and nonmortarMeshId prior to calling this routine.");
+   SLIC_ERROR_IF( this->mortarMeshId == this->nonmortarMeshId,
```
ebchin, line 776: Should the macros in this file be `SLIC_ERROR_ROOT_IF`?

src/tribol/interface/tribol.cpp
srwopschall: Good catch. This should be a warning. Fixed. 

I updated this to be `SLIC_WARNING_ROOT'. If a coupling scheme is invalid on one rank for non-null meshes it should be invalid on all ranks for non-null meshes.

ebchin (commented):
Thanks @srwopschall.  Most comments are regarding clarity and small changes/additons that I thought made sense, but I didn't spot any bugs.  I'll leave my review comments for now, but I'd also like to check this branch on a few serac test problems.

src/tribol/geom/ContactPlane.cpp
srwopschall: Fixed.

src/tribol/geom/ContactPlane.cpp
srwopschall: I changed to `SLIC_DEBUG` which it should be, and elaborated as much as I can on the note. If this does get triggered it will be one of those gross things to debug.

src/tribol/geom/ContactPlane.cpp
srwopschall: I actually need to make this consistent with the 3D case where I reset the area frac. I updated this.

src/tribol/geom/ContactPlane.cpp
srwopschall: Fixed.

src/tests/tribol_mortar_data_weights.cpp
srwopschall: Fixed.

src/tests/tribol_mortar_gap.cpp
srwopschall: Added to tests where appropriate for completeness. Further refactor outside scope of this pr.

src/tribol/common/logger.hpp
srwopschall: I think this makes sense upon further reflection. I have removed it.

src/tribol/geom/ContactPlane.hpp
srwopschall: I don't see where the pair is modified. The contact plane is created, but the pair is a fixed interface pair, at least for this routine. 

We can certainly consider your points (1) and (2) later on.

ebchin: If the interface pair isn't modified, then it should be `const` in the function signature.

src/tests/tribol_tet_mesh.cpp
srwopschall: I removed extraneous ones. Some tests needed it. This is something to investigate further.

src/tribol/geom/ContactPlane.hpp
srwopschall: Unfortunately this will have a cascading effect touching many routines. While we should fix it, it is outside the scope and schedule of this PR. Will make an issue.

src/tribol/mesh/CouplingScheme.hpp
srwopschall: Done.

src/tribol/physics/Mortar.cpp
srwopschall: Fixed.

src/tribol/physics/Mortar.cpp
srwopschall: Fixed.

src/tribol/utils/TestUtils.cpp
srwopschall: Debatable. This test utility class isn't designed to be MPI aware. using 'ROOT' probably is best for clarity, but the use of the macros is fine the way they are technically because this is meant for serial tests. This is outside the scope of this PR, but a good point.

src/tribol/geom/ContactPlane.hpp
srwopschall: Added issue to investigate and correct const correctness in CG routines.

src/tribol/geom/ContactPlane.cpp
srwopschall: Leaving for now. This is deep in the weeds. I made a change that helped logging but broke the cg. This does need to be done.

src/tribol/geom/ContactPlane.cpp
srwopschall: Yes, nonmortar face. Comment updated.

src/tribol/common/Parameters.hpp
srwopschall: This may be deprecated in future logging updates, so going to leave for now.

src/tribol/geom/ContactPlane.hpp
srwopschall: fixed.

src/tribol/geom/ContactPlane.hpp
srwopschall: fixed.

src/tribol/geom/ContactPlane.hpp
srwopschall: fixed.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Yes. This came up in a separate discussion. I think a utility to do this would be very useful.

src/tribol/geom/ContactPlane.hpp
srwopschall: This will be investigated further per new issue; not in this pr. This may be [in] and need to be const.

src/tribol/geom/ContactPlane.hpp
srwopschall: will investigate per issue as to whether it is actually in/out or should be const.

src/tribol/geom/GeomUtilities.cpp
srwopschall: This will be part of a future PR refining how the logging is done. There are a few spots where DEBUG prints are contingent on debug calculations. We need a solid and uniform way to guard these. (future PR)

src/tribol/interface/tribol.cpp
srwopschall: No longer in code.

src/tribol/interface/tribol.cpp
srwopschall: This has been removed. Host code will handle this. We will look to refine host-code/tribol logging and who does what in the future.

src/tribol/interface/tribol.cpp
srwopschall: We may want to do this, but we haven't fully addressed who does what, especially if a host-code doesn't use slic. Will refine in the future.

src/tribol/interface/tribol.cpp
srwopschall: Good point. Fixed.

src/tribol/interface/tribol.cpp
srwopschall: I missed changing these to ROOT macros. Fixed.

src/tribol/interface/tribol.cpp
srwopschall: I changed to debug. This is a good point. If the host-code doesn't guard calls for null data, then this call will be made with a nullptr, which I think is ok and will happen as you stated. In a future PR we will make sure we uniformly handle "null-calls" and associated logging.

srwopschall: As a note, I do set the mesh to invalid for a call to this routine with a null pointer to the gap data. In principle I could check the given mesh to see if it is a null mesh, but alternatively, it could be the host-codes responsibility to only register non-null data. Handling this more gracefully is a future PR and I'm interested in your input from the implicit testing/side of things.

srwopschall: Last thing. I actually changed the if check to only set the mesh to invalid for null pointer with non-null mesh.

src/tribol/interface/tribol.cpp
srwopschall: I changed the if-block. Unfortunately a lot of checks in tribol are predicated upon there being passed in a non-null pointer. A future refactor is to make sure every API call can be made on all ranks, even for null meshes or null data. That is outside this scope.

src/tribol/interface/tribol.cpp
srwopschall: We use this elsewhere so I assume so (?).

src/tribol/interface/tribol.cpp
srwopschall: fixed.

src/tribol/interface/tribol.cpp
srwopschall: This is tricky. Some Tribol API functions call this routine, but a host-code may not call those functions for null-meshes (currently), so not all ranks would get into this call, which probably poses an issue. Going to leave for now since the checks are only for null pointers.

src/tribol/interface/tribol.cpp
srwopschall: This will be part of a bigger refinement PR. See response to previous comment. (i.e. potentially not all ranks call this).

src/tribol/interface/tribol.cpp
srwopschall: Future PR/issue.

src/tribol/interface/tribol.hpp
srwopschall: out of date.

src/tribol/interface/tribol.hpp
srwopschall: Good catch. Fixed.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: I removed it. I think appropriate logging happens at a higher level.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Keep these things in mind for a further refinement PR. I agree we shouldn't error out, but giving the host-code some control of how to proceed will be good.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Leaving for now.

src/tribol/mesh/CouplingScheme.hpp
srwopschall: Added.

src/tribol/utils/ContactPlaneOutput.cpp
srwopschall: This ought to just be reset from invalid to valid? Minor detail.

src/tribol/geom/ContactPlaneManager.cpp
srwopschall: Good catch.

src/tribol/interface/tribol.cpp
ebchin: I don't see the change to SLIC_DEBUG here; is does that still need to be pushed?

ebchin: Never mind, I see the changed if statement!

srwopschall: This got a little more confusing. I added checks for null pointers and null vs. non-null meshes. If null pointer and null mesh, it is DEBUG, if null pointer and non-null mesh it is warning and handled later in the code. Again, there is no uniformity to how register calls are handled with null meshes. That will be a future pr.

ebchin (approved):
Thanks for addressing my comments @srwopschall.  Let's keep the comments from this PR in mind when we have a chance to refine things further.

src/tribol/interface/tribol.cpp
```diff
@@ -692,6 +692,12 @@ void registerMortarGaps( integer meshId,
                     "on mesh " << meshId << ".");
       mesh.m_isValid = false;
    }
+   else if (gaps == nullptr && mesh.m_numCells <= 0)
```
ebchin: Should this be >= 0?

src/tribol/interface/tribol.cpp
```diff
@@ -717,6 +723,12 @@ void registerMortarPressures( integer meshId,
                     "on mesh " << meshId << ".");
       mesh.m_isValid = false;
    }
+   else if (pressures == nullptr && mesh.m_numCells <= 0)
```
ebchin: I think this should be >= 0 also

src/tribol/interface/tribol.cpp
srwopschall: the first check is null pointer + non-null mesh (mesh.m_numCells > 0) and we issue a warning. The else if checks for a null pointer and a null mesh (this should be allowed) (mesh.m_numCells <=0) and issues a debug statement (arguable don't need to issue anything, but might be nice for debug.

src/tribol/interface/tribol.cpp
srwopschall: see comment in registerMortarGaps()

src/tribol/interface/tribol.cpp
ebchin: OK i see it now, I missed that these are new checks (hard to see on GitHub).

srwopschall: this has been resolved per outside discussion.

src/tribol/interface/tribol.cpp
srwopschall: this has been resolved per outside discussion.

src/tribol/interface/tribol.cpp
srwopschall: done.

Issue #39, opened 2023-11-06
`SLIC_DEBUG` and `SLIC_ASSERT` are not active unless `AXOM_DEBUG` is defined


white238:
I was looking through your code and realized you are relying on two sets of SLIC macros that are not enabled. I expect this was an oversight. You should reevaluate which SLIC macros you are using.

For example:

https://github.com/LLNL/Tribol/blob/892d27d21ebce6931d38354f76444c490e8f8042/src/redecomp/MultiRedecomp.cpp#L38

This should probably be a `SLIC_ERROR`.

ebchin:
Fixed with #141

Issue #36, opened 2023-11-01
Investigate and Correct Const Correctness


srwopschall:
Some const correctness issues in computational geometry routines and contact plane routines.

PR #40, opened 2023-11-06
Limit use of Axom CMake targets

white238:
This limits which Axom CMake targets are propagated to downstream users. This should only make `axom::primal` and below required to be propagated, while `axom::quest` is only used internally.

~~I also did a very quick pass-through to quiet all compiler warnings. I won't be offended if you don't like how I fixed them.~~

### DIFF
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -10,7 +10,7 @@ set( contact_examples
      )
 
 
-set(example_depends axom tribol axom::cli11 axom::fmt)
+set(example_depends tribol)
 blt_list_append(TO example_depends ELEMENTS cuda IF ENABLE_CUDA )
 
 foreach( example ${contact_examples} )
@@ -80,7 +80,7 @@ if ( BUILD_REDECOMP )
         NAME       ${example_name}_ex
         SOURCES    ${example}
         OUTPUT_DIR ${EXAMPLE_OUTPUT_DIRECTORY}
-        DEPENDS_ON axom mfem mpi redecomp tribol
+        DEPENDS_ON tribol
         )
 
     if (ENABLE_CUDA)
@@ -107,7 +107,7 @@ if ( BUILD_REDECOMP )
         NAME       ${example_name}_ex
         SOURCES    ${example}
         OUTPUT_DIR ${EXAMPLE_OUTPUT_DIRECTORY}
-        DEPENDS_ON axom mfem mpi redecomp tribol
+        DEPENDS_ON tribol
         )
 
     if (ENABLE_CUDA)
--- a/src/tribol/CMakeLists.txt
+++ b/src/tribol/CMakeLists.txt
@@ -101,7 +101,7 @@ if (ENABLE_FORTRAN)
 endif()
 
 ## setup the dependency list for tribol
-set(tribol_depends axom mfem)
+set(tribol_depends axom::slic mfem)
 blt_list_append(TO tribol_depends ELEMENTS mpi IF TRIBOL_USE_MPI )
 blt_list_append(TO tribol_depends ELEMENTS redecomp IF BUILD_REDECOMP )
 blt_list_append(TO tribol_depends ELEMENTS cuda IF ENABLE_CUDA )
@@ -117,6 +117,9 @@ blt_add_library(
     DEPENDS_ON ${tribol_depends}
     FOLDER tribol
     )
+
+# Don't propagate internal dependencies
+target_link_libraries(tribol PRIVATE axom::quest)
 
 target_include_directories(tribol PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
--- a/src/tribol/CMakeLists.txt
+++ b/src/tribol/CMakeLists.txt
@@ -101,7 +101,7 @@ if (ENABLE_FORTRAN)
 endif()
 
 ## setup the dependency list for tribol
-set(tribol_depends axom::slic mfem)
+set(tribol_depends axom::primal axom::slic mfem)
 blt_list_append(TO tribol_depends ELEMENTS mpi IF TRIBOL_USE_MPI )
 blt_list_append(TO tribol_depends ELEMENTS redecomp IF BUILD_REDECOMP )
 blt_list_append(TO tribol_depends ELEMENTS cuda IF ENABLE_CUDA )
--- a/src/examples/examples_common.hpp
+++ b/src/examples/examples_common.hpp
@@ -8,7 +8,6 @@
 
 // axom includes
 #include "axom/core.hpp"
-#include "axom/mint.hpp"
 #include "axom/primal.hpp"
 #include "axom/slic.hpp"
 #include "axom/CLI11.hpp"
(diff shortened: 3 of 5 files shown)


src/examples/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set( contact_examples
      )
 
 
-set(example_depends axom tribol axom::cli11 axom::fmt)
+set(example_depends tribol)
 blt_list_append(TO example_depends ELEMENTS cuda IF ENABLE_CUDA )
```
ebchin, line 14: Is this one still needed?

src/tribol/CMakeLists.txt
```diff
@@ -118,6 +118,9 @@ blt_add_library(
     FOLDER tribol
     )
 
+# Don't propagate internal dependencies
+target_link_libraries(tribol PRIVATE axom::quest)
```
ebchin, line 122: 👍

src/tribol/physics/CommonPlane.cpp
```diff
@@ -436,12 +436,13 @@ int ApplyNormal< COMMON_PLANE, PENALTY >( CouplingScheme const * cs )
...
-      //SLIC_DEBUG("force sum, side 2, pair " << kp << ": " << dbg_sum_force2 );
-      //SLIC_DEBUG("phi 1 sum: " << phi_sum_1 );
-      //SLIC_DEBUG("phi 2 sum: " << phi_sum_2 );
+      if (logLevel == DEBUG)
+      {
+         SLIC_INFO("force sum, side 1, pair " << kp << ": " << -dbg_sum_force1 );
+         SLIC_INFO("force sum, side 2, pair " << kp << ": " << dbg_sum_force2 );
+         SLIC_INFO("phi 1 sum: " << phi_sum_1 );
+         SLIC_INFO("phi 2 sum: " << phi_sum_2 );
+      }
```
ebchin: I think this should stay commented out for now @srwopschall?  Or potentially removed/moved to common plane tests?

src/tribol/utils/TestUtils.cpp
```diff
@@ -833,7 +833,7 @@ void TestMesh::allocateAndSetVelocities( int meshId, real valX, real valY, real
                   "not a valid mesh id." );
    }
 
-   SLIC_DEBUG_IF( deleteVels, "TestMesh::allocateAndSetVelocities(): " << 
+   SLIC_INFO_IF( deleteVels, "TestMesh::allocateAndSetVelocities(): " <<
```
ebchin: @srwopschall can you verify these changes are OK?  I think `SLIC_DEBUG_IF` seems like the right level here, but #39 will have to be addressed first.

ebchin (approved):
Thanks @white238!  The CMake changes look fine to me, but @srwopschall should verify the changes to the slic macros before merging.

src/examples/CMakeLists.txt
kennyweiss: Yes. Unfortunately, the `cuda` dependency needs to be explicitly listed.

src/tests/tribol_mortar_data_weights.cpp
```diff
@@ -78,8 +78,6 @@ void TestMortarWeights( tribol::CouplingScheme const * cs, double exact_area, do
       mortarMesh.sortSurfaceNodeIds();
    }
 
-   int nodeOffset = mortarMesh.m_sortedSurfaceNodeIds[ mortarMesh.m_numSurfaceNodes-1 ] + 1;
```
kennyweiss: It looks like this variable was used for debugging. @srwopschall -- would you prefer to keep it and the output below as comments?

src/tests/tribol_mortar_force.cpp
```diff
@@ -80,7 +80,7 @@ class MortarForceTest : public ::testing::Test
       real * z = this->z;
 
       // register the mesh with tribol
-      int cellType;
+      int cellType = 0;
```
kennyweiss: Perhaps this should be initialized to `UNDEFINED_ELEMENT`?

https://github.com/LLNL/Tribol/blob/892d27d21ebce6931d38354f76444c490e8f8042/src/tribol/common/Parameters.hpp#L49-L58

(here and elsewhere)

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1119,6 +1120,7 @@ bool SegmentIntersection2D( const real xA1, const real yA1, const real xB1, cons
      SLIC_DEBUG_IF( xDiff > diffTol || yDiff > diffTol,
```
kennyweiss: (One line up) `diffTol` should be a constexpr:
```diff
- real diffTol = 1.0E-3;
+ constexpr real diffTol = 1.0E-3;
```

and perhaps the other vars should be `const`?

src/tribol/physics/CommonPlane.cpp
kennyweiss: I agree that these lines should stay commented out

src/tribol/utils/TestUtils.cpp
kennyweiss: @white238 -- we were observing way too much output in a user application, so @srwopschall recently converted a lot of `INFO` logs to `DEBUG`. He's planning a quick follow-up PR to increase the verbosity where appropriate.

Tribol likely needs a user-controlled log verbosity level for INFO and possibly DEBUG log messages. Perhaps going from 0 (no logs) to 5 (all the logs).

kennyweiss (commented):
Thanks for the fixes @white238!

I think @srwopschall will want to update the changes to the log levels to ensure they're not too verbose. He's already planning a follow-up PR to #29 focus on the log statements.

src/tests/tribol_mortar_data_weights.cpp
srwopschall: Yes, let's keep this variable and the output below that uses it as comments. @white238 can we restore this and comment out?

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1099,6 +1099,7 @@ bool SegmentIntersection2D( const real xA1, const real yA1, const real xB1, cons
    }
 
    // TODO refine how these debug calculations are guarded
+#ifdef AXOM_DEBUG
```
srwopschall: When/how is this defined now? Does this properly address issue https://github.com/LLNL/Tribol/issues/39?

src/tribol/geom/GeomUtilities.cpp
srwopschall: This can be addressed when https://github.com/LLNL/Tribol/issues/36 is addressed.

src/tribol/physics/CommonPlane.cpp
srwopschall: Agree, please leave as commented out. How lines like this are handled with be addressed in future PR.

src/tribol/utils/TestUtils.cpp
srwopschall: [src/tribol/geom/GeomUtilities.cpp](https://github.com/LLNL/Tribol/pull/40/files#diff-881a58c856789d0769a178217fdaa4b9fdd4e1f19e4293fdd5e68f50b07f7b6f) uses AXOM_DEBUG. Can we do that here? Is AXOM_DEBUG properly defined?

white238:
I am going to revert the warnings commit and leave this to you guys to decide how you want the warnings fixed.

srwopschall (approved):

PR #44, opened 2023-11-18
Make forces/gaps dual

ebchin:
This PR changes the behavior of the force and gap fields returned in the MFEM interface to be dual vectors.  In other words, the shared DOFs across processors will need to be summed in parallel to obtain the value at the DOF.

This behavior is required to use the LOR transfer operator and is standard in MFEM.

Note this is a breaking change.

Further, there are a few small fixes in the MFEM interface and redecomp and an LOR test for the common plane method.

### DIFF
--- a/src/examples/mfem_mortar_lm_patch.cpp
+++ b/src/examples/mfem_mortar_lm_patch.cpp
@@ -377,8 +377,8 @@ int main( int argc, char** argv )
   mfem::Vector g;
   tribol::getMfemGap(coupling_scheme_id, g);
 
-  // Apply a restriction operator on the submesh: maps dofs stored in g to tdofs
-  // stored in G for parallel solution of the linear system.
+  // Apply a transpose prolongation operator on the submesh: maps dofs stored in g to tdofs stored in G for parallel
+  // solution of the linear system.
   //
   // NOTE: gap is a dual field so we apply the transpose prolongation operator
   {
--- a/src/examples/mfem_mortar_lm_patch.cpp
+++ b/src/examples/mfem_mortar_lm_patch.cpp
@@ -380,18 +380,15 @@ int main( int argc, char** argv )
   // Apply a restriction operator on the submesh: maps dofs stored in g to tdofs
   // stored in G for parallel solution of the linear system.
   //
-  // NOTE: gap is a dual field, but we still apply R here because tribol stores
-  // this like a ParGridFunction: shared DOFs have the same (summed) value on
-  // all ranks
+  // NOTE: gap is a dual field so we apply the transpose prolongation operator
   {
     auto& G = B_blk.GetBlock(1);
-    auto& R_submesh = *tribol::getMfemPressure(coupling_scheme_id)
-      .ParFESpace()->GetRestrictionMatrix();
-    R_submesh.Mult(g, G);
+    auto& P_submesh = *tribol::getMfemPressure(coupling_scheme_id)
+      .ParFESpace()->GetProlongationMatrix();
+    P_submesh.MultTranspose(g, G);
   }
 
-  // Create a single HypreParMatrix from blocks (useful for different
-  // solvers/preconditioners). This process requires two steps: (1) store
+  // Create a single HypreParMatrix from blocks. This process requires two steps: (1) store
   // pointer to the BlockOperator HypreParMatrixs in a 2D array and (2) call
   // mfem::HypreParMatrixFromBlocks() to create the merged, single
   // HypreParMatrix (without blocks).
--- a/src/redecomp/RedecompMesh.cpp
+++ b/src/redecomp/RedecompMesh.cpp
@@ -296,7 +296,8 @@ void RedecompMesh::BuildRedecomp()
   }
 
   // Finalize mesh topology
-  FinalizeTopology(false);
+  auto generate_boundary = false;
+  FinalizeTopology(generate_boundary);
 
   // Fill r2p_elem_offsets_ with element rank offsets
   // r2p = redecomp to parent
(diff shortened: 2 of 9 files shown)


src/tribol/mesh/MfemData.cpp
```diff
@@ -112,22 +112,41 @@ void SubmeshRedecompTransfer::SubmeshToRedecomp(
...
+    dst_fespace_ptr = submesh_lor_xfer_->GetLORGridFn().ParFESpace();
   }
   // transfer data from redecomp mesh
-  redecomp_xfer_.TransferToParallel(redecomp_src, *dst_ptr);
+  mfem::ParGridFunction dst_gridfn(dst_fespace_ptr, *dst_ptr);
+  redecomp_xfer_.TransferToParallel(redecomp_src, dst_gridfn);
+  // using redecomp, shared dof values are set equal (i.e. a ParGridFunction), but we want the sum of shared dof
+  // values to equal the actual dof value when transferring dual fields (i.e. force and gap) back to the parallel mesh
+  // following MFEMs convention.  set non-owned DOF values to zero.
+  auto P_I = dst_fespace_ptr->Dof_TrueDof_Matrix()->GetDiagMemoryI();
```
srwopschall, line 137: Add a comment on what P_I is.

src/tribol/mesh/MfemData.hpp
```diff
@@ -694,7 +703,7 @@ class MfemMeshData
    */
   integer GetMesh1NE() const
```
srwopschall, line 704: Strongly consider checking to see if the data is there and in fact updated.

src/tribol/mesh/MfemData.hpp
```diff
@@ -739,7 +748,7 @@ class MfemMeshData
    *
    * @return integer 
    */
-  integer GetElemType() const { return elem_type_; }
+  integer GetElemType() const { return update_data_->elem_type_; }
```
srwopschall: Can we return the Tribol element type enum here instead of integer?

src/tribol/mesh/MfemData.hpp
```diff
@@ -780,8 +789,9 @@ class MfemMeshData
   /**
    * @brief Get the nodal response vector on the parent mesh
    *
-   * @param [out] r Pre-allocated, initialized mfem::Vector to which response
-   * vector is added
+   * @note This is stored as a dual vector, meaning the shared DOFs must be summed over all ranks to obtain their value.
```
srwopschall: Consider: This is stored as a mfem dual vector, meaning that they expect the shared DOFs to be summed over all ranks to obtain their value.

src/examples/mfem_mortar_lm_patch.cpp
```diff
@@ -380,18 +380,15 @@ int main( int argc, char** argv )
   // Apply a restriction operator on the submesh: maps dofs stored in g to tdofs
```
srwopschall: Does this comment need to change from using a restriction operator to the prolongation operator?

src/redecomp/RedecompMesh.cpp
```diff
@@ -296,7 +296,7 @@ void RedecompMesh::BuildRedecomp()
   }
 
   // Finalize mesh topology
-  FinalizeTopology();
+  FinalizeTopology(false);
```
srwopschall: Maybe indicate what passing 'false' does here. Or initialize some boolean named something to make this clearer. Is there a case where you'd pass 'true'?

src/tests/tribol_mfem_mortar_lm.cpp
```diff
@@ -194,8 +194,8 @@ class MfemCommonPlaneTest : public testing::TestWithParam<int> {
   // restriction operator on submesh: maps dofs stored in g to tdofs stored in G
```
srwopschall: Change note from restriction operator to prolongation operator?

src/tribol/interface/mfem_tribol.hpp
```diff
@@ -238,16 +237,14 @@ std::unique_ptr<mfem::BlockOperator> getMfemBlockJacobian( integer csId );
 /**
  * @brief Returns gap vector to a given mfem::Vector
  *
- * @pre Coupling scheme cs_id must be registered using
- * registerMfemCouplingScheme()
- * @pre Redecomp mesh must be created and up to date by calling
- * updateMfemParallelDecomposition()
- * @pre Response vector must be up to date for current geometry by calling
- * update()
+ * @note This is stored as a dual vector, meaning the shared DOFs must be summed over all ranks to obtain their value.
```
srwopschall: "This is stored as a mfem dual vector, meaning they expect the shared DOFs to be summed over all ranks to obtain the appropriate values."?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1554,6 +1553,8 @@ class MfemSubmeshData
   /**
    * @brief Get the gap vector on the parent-linked boundary submesh
    *
+   * @note This is stored as a dual vector, meaning the shared DOFs must be summed over all ranks to obtain their value.
```
srwopschall: This is stored as a mfem dual vector, meaning they expect the shared DOFs to be summed over all ranks to obtain the appropriate values. or something like that.

srwopschall (approved):
Excellent catch and fix.

src/tribol/mesh/MfemData.hpp
ebchin: Fixed.  The `GetUpdateData()` method does this check.

src/redecomp/RedecompMesh.cpp
ebchin: Ahh this is to prevent boundary data (elements, attributes) from being created, since 1) it isn't clear what we would even want to be considered the boundary mesh and 2) the boundary mesh isn't used anywhere in Tribol.  Also, it sets the boundary attributes to zero which causes a lot of warnings in the output when true.  I'll name the variable.

ebchin:
Thanks for the review @srwopschall!  All of your comments are addressed with the latest commit.

PR #52, opened 2024-01-10
Fix Axom export set cmake error

chapman39:
Resolves the following error seen in LiDO:

```
CMake Error: install(EXPORT "tribol-targets" ...) includes target "redecomp" which requires target "primal" that is not in this export set, but in multiple other export sets: lib/cmake/axom-targets.cmake, lib/cmake/serac-targets.cmake.
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "primal" target to a single export.
CMake Error: install(EXPORT "tribol-targets" ...) includes target "tribol" which requires target "primal" that is not in this export set, but in multiple other export sets: lib/cmake/axom-targets.cmake, lib/cmake/serac-targets.cmake.
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "primal" target to a single export.
CMake Error: install(EXPORT "tribol-targets" ...) includes target "tribol" which requires target "slic" that is not in this export set, but in multiple other export sets: lib/cmake/axom-targets.cmake, lib/cmake/serac-targets.cmake.
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "slic" target to a single export.
CMake Error: install(EXPORT "tribol-targets" ...) includes target "tribol" which requires target "quest" that is not in this export set, but in multiple other export sets: lib/cmake/axom-targets.cmake, lib/cmake/serac-targets.cmake.
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "quest" target to a single export.
```

I don't fully understand what's going on here. There may be a more correct solution.

In `cmake/SetupThirdParty.cmake`, I've tried appending primal, slic, and quest to `TPL_DEPS` instead, but it simply generated more of the same error.

https://github.com/LLNL/Tribol/blob/develop/cmake/SetupThirdParty.cmake#L33

### DIFF
--- a/cmake/SetupThirdParty.cmake
+++ b/cmake/SetupThirdParty.cmake
@@ -26,7 +26,30 @@ endif()
 #------------------------------------------------------------------------------
 # axom
 #------------------------------------------------------------------------------
-if (DEFINED AXOM_DIR)
+if (TARGET axom)
+    # Case: Tribol included in project that also creates an axom target, no need to recreate axom
+    message(STATUS "Axom support is ON, using existing axom target")
+
+    # Check to make sure the components and dependencies of axom exist
+    # NOTE(chapman39@llnl.gov): Cannot simply install axom, since it is an imported library
+    set(_axom_comp ${AXOM_COMPONENTS_ENABLED};cli11;fmt;sol;lua;hdf5;sparsehash)
+    foreach(_comp ${_axom_tpls})
+      if (NOT TARGET _comp)
+        message(FATAL_ERROR
+          "Axom is a required dependency for tribol. "
+          "${_comp} is a missing component of Axom that is also required.")
+      endif()
+    endforeach()
+    
+    # Add them to this export set but don't prefix it with tribol::
+    install(TARGETS              ${_axom_comp}
+            EXPORT               tribol-targets
+            DESTINATION          lib)
+    unset(_axom_comp)
+
+    set(AXOM_FOUND TRUE CACHE BOOL "" FORCE)
+
+elseif (DEFINED AXOM_DIR)
   message(STATUS "Setting up external Axom TPL...")
   include(${PROJECT_SOURCE_DIR}/cmake/thirdparty/SetupAxom.cmake)
 
--- a/cmake/SetupThirdParty.cmake
+++ b/cmake/SetupThirdParty.cmake
@@ -30,22 +30,14 @@ if (TARGET axom)
     # Case: Tribol included in project that also creates an axom target, no need to recreate axom
     message(STATUS "Axom support is ON, using existing axom target")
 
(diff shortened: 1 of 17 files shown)


ebchin:
The limited Axom targets were introduced in #40, so we'd like to retain those if possible.  @white238, any idea on how to fix the LiDO error without changing Tribol's Axom targets?

chapman39:
I've realized now since I'm testing Tribol as a sub-module in a project that already builds Axom, we need something similar to https://github.com/LLNL/Tribol/pull/51, but for Axom. I'm working on this now.

white238 (approved):

cmake/SetupThirdParty.cmake
```diff
@@ -26,7 +26,30 @@ endif()
...
 # axom
 #------------------------------------------------------------------------------
-if (DEFINED AXOM_DIR)
+if (TARGET axom)
+    # Case: Tribol included in project that also creates an axom target, no need to recreate axom
+    message(STATUS "Axom support is ON, using existing axom target")
+
+    # Check to make sure the components and dependencies of axom exist
+    # NOTE(chapman39@llnl.gov): Cannot simply install axom, since it is an imported library
+    set(_axom_comp ${AXOM_COMPONENTS_ENABLED};cli11;fmt;sol;lua;hdf5;sparsehash)
```
white238: Using this variable from Axom would clean this up and hopefully be more robust.

https://github.com/LLNL/axom/blob/982c6d3c00a607c8600119dd1db847fd5acff5e9/src/cmake/AxomConfig.cmake#L157

kennyweiss: I agree with @white238  that we should use an export variable for the list of targets.

The names deps on this line aren't all required for axom, e.g. `lua`, `sol`, and `hdf5` are optional. I think `sparsehash` is also optional, but it's been a while since I last checked.

chapman39: Should the list be updated from `axom_exported_targets` to something like `AXOM_EXPORTED_TARGETS` now that it will be used outside that cmake section?

chapman39: Update: i ended up needing to use `axom_exported_targets` rather than `_axom_exported_targets`

chapman39: Also, i had to remove openmp from list, since it was included in another export set

From serac codevelop: `CMake Error: install(EXPORT "tribol-targets" ...) includes target "openmp" more than once in the export set.`

white238: That should be fixed when the new BLT is released but good job working around it.

ebchin (approved):

srwopschall (approved):

PR #53, opened 2024-01-10
Mark contact candidates

srwopschall:
- This more clearly marks contact candidates emerging from the computational geometry with associated contact planes
- The contact candidate mark on the interface pairs is more clearly used throughout the code
- An old 'inContact' boolean on the interface pairs is deprecated in favor of marking 'in contact' on the contact plane
- The timestep calculation skips interface pairs that are NOT contact candidates, which is a bugfix.

### DIFF
--- a/src/tribol/geom/ContactPlane.cpp
+++ b/src/tribol/geom/ContactPlane.cpp
@@ -32,6 +32,7 @@ FaceGeomError CheckInterfacePair( InterfacePair& pair,
                                   bool& inContact )
 {
    inContact = false;
+   pair.isContactCandidate = false;
 
    // note: will likely need the ContactCase for specialized 
    // geometry check(s)/routine(s)
@@ -69,6 +70,7 @@ FaceGeomError CheckInterfacePair( InterfacePair& pair,
            }
            else if (cpTemp.m_inContact)
            {
+              pair.isContactCandidate = true;
               cpMgr.addContactPlane( cpTemp ); 
               inContact = true;
            }
@@ -89,6 +91,7 @@ FaceGeomError CheckInterfacePair( InterfacePair& pair,
            }
            else if (cpTemp.m_inContact)
            {
+              pair.isContactCandidate = true;
               cpMgr.addContactPlane( cpTemp ); 
               inContact = true;
            }
@@ -116,6 +119,7 @@ FaceGeomError CheckInterfacePair( InterfacePair& pair,
 
             if (cpTemp.m_inContact)
             {
+               pair.isContactCandidate = true;
                cpMgr.addContactPlane( cpTemp );
                inContact = true;
             }
--- a/src/tribol/geom/ContactPlane.cpp
+++ b/src/tribol/geom/ContactPlane.cpp
@@ -29,10 +29,9 @@ namespace tribol
 FaceGeomError CheckInterfacePair( InterfacePair& pair,
                                   ContactMethod const cMethod,
                                   ContactCase const TRIBOL_UNUSED_PARAM(cCase),
-                                  bool& inContact )
+                                  bool& isInteracting )
 {
-   inContact = false;
-   pair.isContactCandidate = false;
+   isInteracting = false;
 
    // note: will likely need the ContactCase for specialized 
    // geometry check(s)/routine(s)
@@ -66,17 +65,16 @@ FaceGeomError CheckInterfacePair( InterfacePair& pair,
 
            if (face_err != NO_FACE_GEOM_ERROR)
            {
-              inContact = false;
+              isInteracting = false;
            }
            else if (cpTemp.m_inContact)
            {
-              pair.isContactCandidate = true;
               cpMgr.addContactPlane( cpTemp ); 
-              inContact = true;
+              isInteracting = true;
            }
            else
            {
-              inContact = false;
+              isInteracting = false;
            }
            return face_err;
          }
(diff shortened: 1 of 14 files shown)


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -916,23 +916,23 @@ int CouplingScheme::apply( integer cycle, real t, real &dt )
...
      else
      {
-        pair.inContact = true;
+        pair.isContactCandidate = true;
         ++numActivePairs;
      }
      
      // update the InterfacePairs container on the coupling scheme 
-     // to reflect the change to "in-contact"
+     // to reflect the change to isContactCandidate = true
```
liu15: Couldn't this be set to false on 926?

srwopschall: Yes. The comment has been updated.

liu15 (approved):

src/tribol/geom/ContactPlane.cpp
```diff
@@ -29,9 +29,9 @@ namespace tribol
 FaceGeomError CheckInterfacePair( InterfacePair& pair,
                                   ContactMethod const cMethod,
                                   ContactCase const TRIBOL_UNUSED_PARAM(cCase),
-                                  bool& inContact )
+                                  bool& isInteracting )
```
ebchin, line 32: I think the header needs to be updated too.  Also, a definition of an interacting face pair would be helpful.

src/tribol/interface/tribol.cpp
```diff
@@ -947,14 +947,19 @@ void setInterfacePairs( integer couplingSchemeIndex,
       // since a user may specify interface pairs that aren't, nor should be in contact.
       InterfacePair pair { meshId1[i], pairType1[i], pairIndex1[i],
                            meshId2[i], pairType2[i], pairIndex2[i],
-                           false, i };
+                           true, i };
```
ebchin: Why is this changed from false to true?  Should we use a constructor that doesn't set this if this is just a placeholder value?

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1372,7 +1382,7 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
       real max_delta1 = proj_ratio * t1;
       real max_delta2 = proj_ratio * t2;
 
-      if (pair.inContact) // pair passes all geometric checks
+      if (cpMgr.m_inContact[cpID]) // pair passes all geometric checks
```
ebchin: Is the cpMgr the only place where a pair's status is stored now?

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -509,16 +514,21 @@ class GridSearch
...
           {
+             pair.isContactCandidate = true;
              pair.pairId = k;
              contactPairs->addInterfacePair( pair );
              ++k;
           }
+          else
+          {
+             pair.isContactCandidate = false;
+          }
```
ebchin: Seems like this block is pretty similar to the above one (L435-449).  Can they be moved to a single function?  Maybe a new InterfacePair constructor?

ebchin (changes requested):
Thanks @srwopschall, these are good changes that make the code clearer.  I left a few comments, but the main one that needs to be addressed is the ContactPlane header.

src/tribol/search/InterfacePairFinder.cpp
srwopschall: We have to check the return value of the geomFilter routine to determine whether to add the interface pair to the manager or not. I moved setting pair.isContactCandidate inside geomFilter (since geomFilter is in fact responsible for performing initial checks on valid face-pairs), which simplifies the lines you point out here. I think this is cleaner?

src/tribol/geom/ContactPlane.hpp
```diff
@@ -30,7 +30,10 @@ class ContactPlaneManager;
  * \param [in] pair interface pair containing pair related indices
  * \param [in] cMethod the Tribol contact method
  * \param [in] cCase the Tribol contact Case
- * \param [in] inContact true if pair are in contact per CG routines
+ * \param [in] isInteracting true if pair passes all computational geometry filters 
+ *
+ * \note isInteracting is true indicating a contact candidate for intersecting or 
+ *       nearly intersecting face-pairs with a positive area of overlap
```
ebchin, line 36: Thanks!

src/tribol/mesh/InterfacePairs.hpp
```diff
@@ -51,11 +60,17 @@ struct InterfacePair
...
-   // boolean to hold whether pair is in contact or not in contact
-   bool inContact;
+   // boolean indicating if a binned pair is a contact candidate.
+   // A contact candidate is defined as a face-pair that is deemed geometrically proximate 
+   // by the binning coarse search, and one that passes the finer computational geometry 
+   // (CG) filter/checks. These finer checks identify face-pairs that are intersecting or 
+   // nearly intersecting with positive areas of overlap. These checks do not indicate 
+   // whether a face-pair is contacting, since the definition of 'contacting' is specific 
+   // to a particular contact method and its enforced contstraints. Rather, the CG filter
+   // identifies contact candidacy, or face-pairs likely in contact.
```
ebchin, line 70: This is helpful, too

src/tribol/search/InterfacePairFinder.cpp
ebchin: Yes, much cleaner now.

ebchin (approved):
Looks good!

PR #55, opened 2024-01-26
clarify the output blocks of the jacobian

ebchin:
Adds documentation for what the Jacobian blocks are when using the MFEM Tribol interface.

### DIFF
--- a/src/tribol/interface/mfem_tribol.hpp
+++ b/src/tribol/interface/mfem_tribol.hpp
@@ -217,17 +217,21 @@ void getMfemResponse( integer cs_id, mfem::Vector& r );
 /**
  * @brief Get assembled contact contributions for the Jacobian matrix
  *
- * @pre Coupling scheme cs_id must be registered using
- * registerMfemCouplingScheme()
- * @pre Redecomp mesh must be created and up to date by calling
- * updateMfemParallelDecomposition()
+ * @pre Coupling scheme cs_id must be registered using registerMfemCouplingScheme()
+ * @pre Redecomp mesh must be created and up to date by calling updateMfemParallelDecomposition()
  * @pre Tribol data must be up to date for current geometry by calling update()
  *
- * This method requires registration of an mfem::ParMesh with the coupling
- * scheme. The Jacobian contributions are split into a 2x2 block structure. The
- * first row (or column) block is associated with the coordinate grid function
- * degrees of freedom and the second row (or column) block is associated with
- * the Lagrange multiplier degrees of freedom.
+ * This method requires registration of an mfem::ParMesh with the coupling scheme. The Jacobian contributions are split
+ * into a 2x2 block structure. The first row (or column) block is associated with the coordinate grid function degrees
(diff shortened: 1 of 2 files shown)


src/tribol/interface/mfem_tribol.hpp
```diff
@@ -217,17 +217,21 @@ void getMfemResponse( integer cs_id, mfem::Vector& r );
...
+ * This method requires registration of an mfem::ParMesh with the coupling scheme. The Jacobian contributions are split
+ * into a 2x2 block structure. The first row (or column) block is associated with the coordinate grid function degrees
+ * of freedom and the second row (or column) block is associated with the Lagrange multiplier (pressure ) degrees of
+ * freedom. Specifically, the blocks are as follows (f = force, u = displacement, g = gap, p = pressure):
+ *
+ *   df/du | df/dp
+ *  ---------------
+ *   dg/du | dg/dp
+ *
+ * Gap/pressure degrees of freedom which do not have a Lagrange multiplier contain empty rows of dg/du and empty columns
```
srwopschall: Is it worth elaborating on why a gap/pressure degree of freedom may exist, but not actually have a Lagrange multiplier? A mention to keeping those dofs around, but not being in the current active set?

src/tribol/mesh/MfemData.cpp
```diff
@@ -968,7 +968,7 @@ std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian(
 {
   // 0 = displacement DOFs, 1 = lagrange multiplier DOFs
   // (0,0) block is empty (for now using SINGLE_MORTAR with approximate tangent)
-  // (1,1) block is empty
+  // (1,1) block id a diagonal matrix with ones on the diagonal of inactive dofs
```
srwopschall: "is a diagonal matrix..." ?

srwopschall (approved):
@ebchin - Thanks for the further elaboration. IMO you can't overdo the amount of comments on this stuff since it can get rather complex. I left two comments for you to address and/or think about. Otherwise, looks good.

ebchin:
@srwopschall, thanks for the review.  Can you check the updated docs address your comments before I merge?

src/tribol/interface/mfem_tribol.hpp
srwopschall:
```suggestion
Note, for contact methods with Lagrange multiplier constraint enforcement, Tribol will have contact contribution placeholders for nodes on both contact surfaces. For instance, in a mortar method, both mortar and nonmortar nodes will have placeholders in the contact contribution data structures. If, however, SINGLE_MORTAR is used as the ContactMethod, the gap/pressure degrees of freedom are ONLY on the nonmortar surface. As a result, Tribol will return Jacobian blocks that have empty rows of dg/du and empty columns of df/dp and ones on the diagonal of the dg/dp block for all mortar node placeholders.
```

srwopschall (approved):
@ebchin - Looks great. I think we clarified the comment about non LM contributions.

PR #57, opened 2024-01-31
Modifies constant penalty calculation to provide equivalence with element-wise calculation

srwopschall:
- The constant penalty calculation was not commensurate with the 'model' used in the element-wise penalty calculation
- This PR modifies the constant penalty calculation to follow the spring-in-series model used in the element-wise calculation.
- This allows a user or developer to compare behavior when using the two penalty types under the condition that each mesh's penalty stiffness is equivalent across penalty types. (E.g. bulk_mod/t = 1, constant_penalty = 1).

### DIFF
--- a/src/tests/tribol_mfem_common_plane.cpp
+++ b/src/tests/tribol_mfem_common_plane.cpp
@@ -60,8 +60,9 @@ protected:
     double lambda = 100000.0;
     // lame parameter (shear modulus)
     double mu = 100000.0;
-    // kinematic penalty
-    double p_kine = (lambda + 2.0 / 3.0 * mu) * std::pow(2.0, ref_levels - 1);
+    // kinematic constant penalty stiffness equivalent to the element-wise calculation, 
+    // which is bulk-modulus over element thickness.
+    double p_kine = (lambda + 2.0 / 3.0 * mu) / (1.0 / std::pow(2.0, ref_levels));
 
     // fixed options
     // location of mesh file. TRIBOL_REPO_DIR is defined in tribol/config.hpp
--- a/src/tribol/physics/CommonPlane.cpp
+++ b/src/tribol/physics/CommonPlane.cpp
@@ -194,6 +194,8 @@ int ApplyNormal< COMMON_PLANE, PENALTY >( CouplingScheme const * cs )
    // loop over interface pairs //
    ///////////////////////////////
    int cpID = 0;
+   int err = 0;
+   bool neg_thickness {false};
    for (IndexType kp = 0; kp < numPairs; ++kp)
    {
       InterfacePair pair = pairs->getInterfacePair(kp);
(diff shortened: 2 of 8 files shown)


liu15 (approved):

src/tests/tribol_mfem_common_plane.cpp
```diff
@@ -60,8 +60,9 @@ class MfemCommonPlaneTest : public testing::TestWithParam<std::pair<int, tribol:
     double lambda = 100000.0;
     // lame parameter (shear modulus)
     double mu = 100000.0;
-    // kinematic penalty
-    double p_kine = (lambda + 2.0 / 3.0 * mu) * std::pow(2.0, ref_levels - 1);
+    // kinematic constant penalty stiffness equivalent to the element-wise calculation, 
+    // which is bulk-modulus over element thickness.
+    double p_kine = (lambda + 2.0 / 3.0 * mu) / (1.0 / std::pow(2.0, ref_levels));
```
ebchin, line 65: Can you make the change in the corresponding example?  https://github.com/LLNL/Tribol/blob/develop/src/examples/mfem_common_plane.cpp

src/tribol/physics/CommonPlane.cpp
```diff
@@ -247,22 +241,24 @@ int ApplyNormal< COMMON_PLANE, PENALTY >( CouplingScheme const * cs )
...
-            // multiply the material modulus (i.e. material stiffness) by each mesh's penalty scale
-            penalty_stiff_per_area = ComputePenaltyStiffnessPerArea( 
-                                        pen_scale1 * mesh1.m_elemData.m_mat_mod[ index1 ],
-                                        mesh1.m_elemData.m_thickness[ index1 ],
-                                        pen_scale2 * mesh2.m_elemData.m_mat_mod[ index2 ],
-                                        mesh2.m_elemData.m_thickness[ index2 ],
-                                        pen_enfrc_options.tiny_length
-                                                                   );
+            // add tiny_length to element thickness to avoid division by zero
+            auto t1 = mesh1.m_elemData.m_thickness[ index1 ] + pen_enfrc_options.tiny_length;
```
ebchin, line 256: Not sure if it's checked elsewhere, but would it be worth it to add a check for negative thicknesses?

ebchin (approved):
Looks good @srwopschall.  Please make sure to update the example before merging.

src/tribol/physics/CommonPlane.cpp
srwopschall: I added a check to ApplyNormal<>. If any face has a negative thickness the routine will return an error at the end of the physics loop after printing a debug message.

src/tests/tribol_mfem_common_plane.cpp
srwopschall: Done.

PR #58, opened 2024-02-05
Modified logging level enum to be a class for clarity.

srwopschall:
- Changing LoggingLevel enum to be a class
- Fix MACRO incompatibility with host-codes
- Adds clarity when using the logging enum

### DIFF
--- a/src/tribol/common/Parameters.hpp
+++ b/src/tribol/common/Parameters.hpp
@@ -33,7 +33,7 @@ constexpr integer ANY_MESH = -1;
 /*!
  * \brief Enumerates the logging level options
  */
-enum LoggingLevel
+enum class LoggingLevel
 {
    UNDEFINED, ///! Undefined 
    DEBUG,     ///! Debug and higher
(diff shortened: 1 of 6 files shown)


kennyweiss (approved):
Thanks @srwopschall! 
It might also be a good idea to note this in the RELEASE-NOTES.

src/tribol/common/Parameters.hpp
```diff
@@ -33,7 +33,7 @@ constexpr integer ANY_MESH = -1;
 /*!
  * \brief Enumerates the logging level options
  */
-enum LoggingLevel
+enum class LoggingLevel
```
liu15: This change does not fix the compilation if DEBUG is set as a macro (the definition expands to

enum class LoggingLevel
{
UNDEFINED,
1,
...
}

liu15: Perhaps use the same case as axom?

srwopschall: I understand the issue now. I appended TRIBOL to each enum value/name and removed the class. I'm not a huge fan of this fix, but the logging needs further refactoring so this may in fact change in the future, and should address your compilation issue.

liu15 (approved):

PR #59, opened 2024-02-06
Guard install

white238:
In older (?) configurations, `mfem` is not namespaced from Axom (I think but its the only place i can think of it coming from).

This will guard the install from erroring out.

### DIFF
--- a/cmake/SetupThirdParty.cmake
+++ b/cmake/SetupThirdParty.cmake
@@ -63,10 +63,16 @@ if (TARGET mfem)
     # Note - white238: I can't seem to get this to pass install testing due to mfem being included
     # in multiple export sets
     message(STATUS "MFEM support is ON, using existing mfem target")
+
     # Add it to this export set but don't prefix it with tribol::
-    install(TARGETS              mfem
-            EXPORT               tribol-targets
-            DESTINATION          lib)
+    # NOTE: imported targets cannot be part of an export set
+    get_target_property(_is_imported mfem IMPORTED)
+    if(NOT "${_is_imported}")
+        install(TARGETS              mfem
+                EXPORT               tribol-targets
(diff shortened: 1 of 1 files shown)


srwopschall (approved):

liu15 (approved):
This worked for us!

cmake/SetupThirdParty.cmake
```diff
@@ -63,10 +63,16 @@ if (TARGET mfem)
     # Note - white238: I can't seem to get this to pass install testing due to mfem being included
     # in multiple export sets
     message(STATUS "MFEM support is ON, using existing mfem target")
+
     # Add it to this export set but don't prefix it with tribol::
-    install(TARGETS              mfem
-            EXPORT               tribol-targets
-            DESTINATION          lib)
+    # NOTE: imported targets cannot be part of an export set
+    get_target_property(_is_imported mfem IMPORTED)
```
liu15, line 69: should this pattern be used for axom as well?

white238: Already done.. though this shouldn't happen again with several improvements. I don't even think it happens on Axom develop atm.

https://github.com/LLNL/axom/pull/1270

liu15: Actually, I meant in the import of axom in this file (above).... in the case that someone is trying to import axom targets?

kennyweiss (approved):

cmake/SetupThirdParty.cmake
white238: When I wrote this fix, I was unable to reproduce this in Axom's develop testing against our install example. `mfem` has been namespaced as `axom::mfem` for a while now.

This just adds a guard against some combinations of library releases.

PR #61, opened 2024-02-08
Fix normal calc for small elements

ebchin:
This PR makes the normal computation work for small elements by reducing the size of the allowed magnitude of the normal to around machine precision.  If the magnitude of the normal is smaller than this (including zero or negative), the routine will now produce a warning and return false.

Additionally, the scope of some variables has been reduced and they are now initialized upon creation.


srwopschall (approved):
@ebchin - Looks good. This should do it.

PR #63, opened 2024-02-09
Mesh Data Calcs After Valid Mesh/Scheme Checks

srwopschall:
- This moves the computeFaceData() into the couplingScheme::init() calculation. 
- This fixes possible bug for host-codes that only call registerMesh() once, which is where the face data was being computed
- This now allocates and computes mesh face data each time tribol::update() is called, which is appropriate, and only does so for valid meshes/schemes, which is also appropriate.
- Tests that just checked mesh data now have to explicitly call mesh.computeFaceData(), as opposed to simply calling tribol::registerMesh().

### DIFF
--- a/src/tribol/interface/tribol.cpp
+++ b/src/tribol/interface/tribol.cpp
@@ -375,6 +375,7 @@ void registerMesh( integer meshId,
    // since Tribol supports null meshes. This is not uncommon in parallel 
    // contact simulations
    mesh.m_meshId = meshId;
+   mesh.m_dim = dim;
    mesh.m_positionX = x;
    mesh.m_positionY = y;
    mesh.m_positionZ = z;
@@ -420,9 +421,6 @@ void registerMesh( integer meshId,
    {
       mesh.sortSurfaceNodeIds();
    }
-
-   mesh.m_dim = dim;
-   mesh.deallocateArrays();
 
    if (mesh.m_numCells > 0)
    {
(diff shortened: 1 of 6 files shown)


src/tribol/interface/tribol.cpp
```diff
@@ -421,28 +421,7 @@ void registerMesh( integer meshId,
...
-      initRealArray( mesh.m_cX,   mesh.m_numCells, 0. );
-      initRealArray( mesh.m_cY,   mesh.m_numCells, 0. );
-      initRealArray( mesh.m_area, mesh.m_numCells, 0. );
-   }
-
-   if (dim == 3 && mesh.m_numCells > 0)
-   {
-      initRealArray( mesh.m_nZ, mesh.m_numCells, 0. );
-      initRealArray( mesh.m_cZ, mesh.m_numCells, 0. );
-   }
```
ebchin: Does this need to be moved?  I think the number of cells and dimension must remain constant once the mesh is registered?  Also, I think these values will be computed for all cells, so there isn't any chance they will be stale if we don't deallocate/reallocate.

srwopschall: that's a good point. If the mesh is registered once, then we don't want the allocation/deallocation to occur every time update is called. I can move these back.

srwopschall: I just pushed the change moving the allocation calls back to registerMesh().

ebchin (approved):

PR #66, opened 2024-02-29
Make timestep vote optional

srwopschall:
- Control use of timestep vote through API
- Refactor timestep vote logging
- Bugfix resulting in negative timestep votes.


srwopschall:
Added @rieben as reviewer for awareness since PR https://github.com/LLNL/Tribol/pull/69 branches off this branch.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1458,74 +1509,115 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
...
-      // of one face's projected overlap-centroid in the 'thickness-direction' 
-      // of the other face (with whom in may be in contact currently, or in 
-      // a velocity projected sense).
-      real proj_delta_n_1 = proj_delta_x1 * fn2[0] + proj_delta_y1 * fn2[1];
-      real proj_delta_n_2 = proj_delta_x2 * fn1[0] + proj_delta_y2 * fn1[1];
-
-      if (dim == 3)
+      // check for case where both faces have nearly zero velocity. If so,
+      // don't compute a velocity projection dt estimate
+      real tiny_vel_proj = 1.e-8;
```
liu15: I know you just moved this, but this dimensional constant is a bit concerning.

src/tribol/common/Parameters.hpp
```diff
@@ -493,6 +493,7 @@ struct parameters_t
   int vis_cycle_incr;            ///! Frequency for visualizations dumps
   VisType vis_type;              ///! Type of interface physics visualization output
   std::string output_directory;  ///! Output directory for visualization dumps
+  bool enable_timestep_vote;     ///! True if host-code desires the timestep vote to be calculated and returned
```
liu15, line 496: could the host code instead provide the alpha multiplier for the timestep vote? (0 for disabled)

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Per our discussion, I removed this check. There are additional complexities to doing this correctly. By removing this check, we simply compute the velocity projected dt estimate for all interface pairs, even those with 'nearly' zero velocity. This change was made for simplicity and clarity.

src/tribol/common/Parameters.hpp
srwopschall: This has not been sufficiently tested to direct a user to what an adequate alpha parameter should be. I think I introduced this parameter to introduce some conservatism to the calculation, but per our discussions, even this is complex because we don't know where that stability limit is. Alpha=1.0 could very well be adequate for some problems, but perhaps introducing some small conservatism is a good idea?

In any case, I think for now introducing the API function controlling this boolean is a little clearer than enable/disabling the timestep calculation inside the timestep function with a scalar parameter. I could see exposing alpha in the future once we have more testing, in addition to the sort of global enabling/disabling of the timestep function in its entirety. 

Thoughts?

liu15: The bool is fine... I was actually thinking about the timestep_pen_frac, which looks like it is already exposed.  I think alpha should be set to 1.0 rather than 0.75.

src/tests/tribol_timestep_vote.cpp
```diff
@@ -302,14 +476,15 @@ TEST_F( CommonPlaneTest, large_velocity_small_gap )
...
+   parameters.enable_timestep_vote = true;
 
    int test_mesh_update_err = 
       this->m_mesh.tribolSetupAndUpdate( tribol::COMMON_PLANE, tribol::PENALTY, 
                                          tribol::FRICTIONLESS, false, parameters );
 
    EXPECT_EQ( test_mesh_update_err, 0 );
 
-   real dt_diff = std::abs(parameters.dt - 0.000747973);
+   real dt_diff = std::abs(parameters.dt - 0.000997297);
```
kennyweiss, line 487: Where does this number come from? Is it empirically obtained (e.g. from running the example)?

kennyweiss (approved):

src/tests/tribol_timestep_vote.cpp
srwopschall: Yes, unfortunately it is empirically obtained by running the example and determining what the computed timestep vote is for this set up inputs.

srwopschall:
@liu15 - I think this PR is ready to go. I have approval by @kennyweiss , but did you have any additional comments or concerns?

liu15 (approved):

PR #69, opened 2024-03-07
Auto-Contact Proximity Checks for Thin Bodies

srwopschall:
- Repurposes the contact_pen_frac to define a maximum allowable interpenetration as a percent of element thickness               between faces in a face-pair when using auto contact
- If face-pairs exceed this maximum allowable (when using auto contact) they are no longer considered contact candidates
- The logic introduced for this is meant to address issues with thin structures/plates when using auto contact when faces on opposite sides of a thin body 'appear' to be in contact to Tribol when they should not be considered.
- Element thicknesses are now required to be registered when using auto contact, regardless of whether the kinematic element penalty is used. The element thicknesses are used to determine max allowable interpen to avoid the thin body issue.
- Coupling scheme tests were added/modified to test and/or properly account for auto contact.
- Computational geometry tests were added to test the maximum allowable interpen when using auto contact.


src/tests/tribol_comp_geom.cpp
```diff
@@ -510,11 +510,16 @@ TEST_F( CompGeomTest, codirectional_normals_3d )
    int numVerts = 4;
    int numCells = 2;
    int lengthNodalData = numCells * numVerts;
-   real element_thickness = 1.;
+   real element_thickness[numCells];
    real x[lengthNodalData];
    real y[lengthNodalData];
    real z[lengthNodalData];
```
kennyweiss: I think the `int` variables need to be `const` or `constexpr` or else the arrays have variable-length, which is non-standard

src/tests/tribol_comp_geom.cpp
```diff
@@ -615,6 +620,273 @@ TEST_F( CompGeomTest, codirectional_normals_3d )
...
+   delete [] fz;
+   delete [] vx;
+   delete [] vy;
+   delete [] vz;
+}
+
+TEST_F( CompGeomTest, auto_contact_gt_max_interpen )
+{
+   // This test uses auto-contact and checks that the face-pair
+   // is included as a conatct candidate, and is in fact in contact
```
kennyweiss: Typo:
```suggestion
   // is included as a contact candidate, and is in fact in contact 
```

kennyweiss (approved):
Thanks @srwopschall !

src/tests/tribol_comp_geom.cpp
srwopschall: I added constexpr to these ints and elsewhere as required in this test. I won't address this in all tests in this PR.

src/tests/tribol_comp_geom.cpp
srwopschall: Good catch. Fixed.

srwopschall:
@liu15 - I think this PR is pretty much ready to go. Do you have any additional comments or questions?

liu15 (approved):

PR #70, opened 2024-03-20
Update How Face Geometry 'Errors' Are Handled

srwopschall:
- Removed erroring out if a face geometry 'error' is detected; there are many instances where we want to punt on CG issues.
- Added a little more granularity to the face geom enum
- Added very basic face geom issue counters on the coupling scheme to give user idea of how many face-issues compared to total number of faces (or face-pairs) have been detected.


srwopschall:
@kennyweiss and @liu15 - This PR resolves the issue of computational geometry 'issues' causing tribol to return an error. Please review at your earliest convenience as multiple codes are experiencing issues related to this. The rationale behind this 'fix' is as follows:

The main thing that this PR does is removes tribol::update() from returning an error if computational geometry issues are encountered. The thinking is that we don't want even just one bad face-pair bringing down the whole simulation. Additionally, there may simply be nasty face geometry that the contact algorithm realistically wants to punt on. We have seen globally correct behavior in the presence of such 'bad' face-pairs. 

The second thing this PR does is a simple, maybe crude attempt at debug reporting what percentage of face-pairs encountered some comp. geom. issue, just to give the developer/tester some idea of how much of a problem these 'issues' may be. This is not a perfect solution, and a more sophisticated system for reporting comp. geom issues (even for catching bugs) is likely a warranted future development task (with careful thought and consideration). That said, it is extremely cumbersome to hunt down how many face-pairs (and dare I say which ones) experience some issue in the CG routines. This will at least give a user/developer an idea of how prevelant any CG issues may be, qualitatively guiding them to either ignore, or pursue a solution, or even attempt to debug the issue.

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -15,6 +15,18 @@
 
 namespace tribol
 {
+// Struct to hold on-rank coupling scheme face-pair reporting data
+// generated from computational geometry issues
+struct PairReportingData
+{
+public:
+   PairReportingData() {};
+   ~PairReportingData() {};
```
kennyweiss: Suggestion (minor): This empty constructor and destructor should be removed. 
The destructor's existence removes the default move constructor.

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -140,6 +152,8 @@ class CouplingScheme
   CouplingSchemeErrors& getCouplingSchemeErrors() { return m_couplingSchemeErrors; }
   CouplingSchemeInfo&   getCouplingSchemeInfo()   { return m_couplingSchemeInfo; }
 
+  PairReportingData& getPairReportingData() {return m_pairReportingData; }
```
kennyweiss: Should this be const?

```suggestion
  const PairReportingData& getPairReportingData() const {return m_pairReportingData; }
```

or can the caller modify this?

kennyweiss (approved):

src/tribol/mesh/CouplingScheme.hpp
srwopschall: Removed. There are a few other enums used for coupling scheme reporting/printing that have empty constructors and destructors. Mind if I remove those too (just those used by the coupling scheme to keep these mods under control)?

src/tribol/mesh/CouplingScheme.hpp
srwopschall: Caller should not modify. Question: the intent of the pair reporting data is that really only the coupling scheme should have control/access to this struct. As is, this getter is not used. Under this design assumption, would you recommend removing it?

src/tribol/mesh/CouplingScheme.hpp
kennyweiss: go for it!

src/tribol/mesh/CouplingScheme.hpp
kennyweiss: In that case, we might as well remove this function and at it back later if needed.

srwopschall: Fixed.

liu15 (approved):

PR #72, opened 2024-03-21
Exposed timestep interpen frac in API.

srwopschall:
Add API function for user to set the timestep interpen fraction.


srwopschall:
@liu15 - small PR to expose that timestep parameter.

liu15 (approved):

PR #73, opened 2024-03-22
Modify API to more clearly support setting of auto contact scale

srwopschall:
- Changed setContactPenFrac(), which was not being used correctly, to setAutoContactPenScale(), which more clearly indicates using a scale factor to scale the element thickness to be used as an auto-contact length scale.
- The auto contact length scale precludes contact on opposite faces of thin plates.

### DIFF
--- a/src/tribol/interface/tribol.cpp
+++ b/src/tribol/interface/tribol.cpp
@@ -67,7 +67,7 @@ void set_defaults()
    parameters.gap_tied_tol                 = 0.1;    // tolerance for how much separation can occur before opposing faces are let go
    parameters.len_collapse_ratio           = 1.E-8;
    parameters.projection_ratio             = 1.E-10;
-   parameters.contact_pen_frac             = 1.0;    // max allowable interpenetration as percent of element thickness for contact candidacy 
+   parameters.auto_contact_pen_frac        = 0.95;   // max allowable interpenetration as percent of element thickness for contact candidacy 
    parameters.timestep_pen_frac            = 3.e-1;  // max allowable interpenetration as percent of element thickness prior to triggering timestep vote (not exposed to API) 
    parameters.enable_timestep_vote         = false;  // true if host-code wants to receive tribol timestep vote
    
@@ -78,7 +78,6 @@ void set_defaults()
    // constituent face elements, then we don't consider the face-pair a contact candidate.
    // Note, auto-contact will require registration of element thicknesses.
    parameters.auto_interpen_check           = false; // true if the auto-contact interpenetration check is used for interpenetrating face-pairs.
-   parameters.auto_contact_len_scale_factor = 1.0;   // sacle factor applied to element thickness for auto contact length scale
 
 }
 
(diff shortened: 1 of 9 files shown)


src/tribol/interface/tribol.cpp
```diff
@@ -187,18 +187,18 @@ void setRatePercentPenalty( int meshId, double r_p )
...
 //------------------------------------------------------------------------------
-void setContactPenFrac( double frac )
+void setAutoContactPenScale( double scale )
 {
    parameters_t & parameters = parameters_t::getInstance();
-   if (frac <= 0.)
+   if (scale <= .95)
    {
-      // Don't set the contact_pen_frac. This will use default
+      parameters.auto_contact_pen_frac = 0.95;
```
liu15: why is this hard coded floor here?

srwopschall: Happy to discuss what this "should" be. The idea is that auto-contact considers penetrating face-pairs whose face-centroids are some scale factor (auto_contact_pen_frac) * the smallest element thickness between the two faces in that face-pair. This serves as the auto-contact length scale to prevent contact between faces on opposite sides of bodies, particularly for thin structures/plates. So, if a thin structure is _one element_ through the thickness, then an auto_contact_pen_frac = 1.0, for instance, risks including those two faces, which is unphysical. To address this "lower bound" case, I set the scale factor to 95% (of the element thickness) here. I'm open to other suggestions. 

The other end of this scenario is that you have a thin structure with _many_ elements through the thickness. In that case, an auto_contact_pen_frac = 0.95 may be too restrictive and not allow enough interpenetration. What I mean by that is that a user may not be concerned about detecting contact between faces on opposite sides of the plate, but for legitimate faces that pass through one another, we may want them to experience a greater amount of interpenetration and desire to use an auto_contact_pen_frac > 0.95. A user would have to know that they can increase this scale factor depending on the number of elements through the thickness, which would then allow for a softer or more forgiving contact response without 'missing' the contact. 

As a note, for now, the most straightforward thing to base the auto-contact length scale off was the element thickness. If a host-code registers the element thickness once at the beginning of the simulation, then this length scale may be too restrictive (i.e. not allow for enough overlap) for possible large distortion. That said, for very large distortion, this potentially too restrictive auto-contact length scale may resulting in 'missing' face-pairs that maybe should have been punted on anyway. More testing will be needed. Alternatively, this may motivate updating the element thickness each cycle so we see the latest current configuration element thicknesses. In practice, however, this may be a moot point if one meshes thin structures with a few or more elements through the thickness.

liu15: I'm not suggesting that > 0.95 is a good idea, but I would suggest that if a user is actually setting that, to allow it (or allow the host code to complain if this is not acceptable).

srwopschall: > 0.95 may be desired if a user wants to let a little more interpen occur. < "some scale factor" and Tribol risks missing all of the contact.  I simply set it to 95% element thickness. This could be 50% element thickness. This is our choice. As the scale factor approaches zero, you increase the risk of missing contact altogether, so that is just what I'm trying to avoid simply in case a user sets it to an "inappropriate" value.

Rather than trying to "save a user from themselves" here, would you suggest simply checking for strict positivity and if the scale factor < 1.0 simply issuing a SLIC_INFO_ROOT to notify the host-code that they _may_ 'miss' some contact?

liu15: Yeah, though I don't know what SLIC_INFO_ROOT is... I would suggest WARNING or INFO level.

srwopschall: Just printing the appropriate level of message on rank 0. I'll go ahead and remove setting this floor for "ill-advised" input values.

srwopschall: The only apprehension I have about this is that from the computational geometry standpoint I know which values I would never use if I were to not expose this to a user because missing contact would be too high of a likely hood. Just to confirm, here we set it to whatever the user desires, and if it is negative, we simply issue a warning....but the simulation will continue even with bad contact behavior. I guess I set the floor to prevent unnecessarily bad contact behavior.

On the other hand I'll add a note to the header to indicate a reasonable value for this parameter so a user will have guidance.

liu15 (approved):

PR #76, opened 2024-04-12
Update types in Tribol

### LINKED ISSUE #12: Double vs. Real
Tribol real covers use case of single and double precision.
Standardize use of 'real' through out code OR
Use C++ double ignoring need for single precision.

ebchin:
This PR introduces new definitions for types in Tribol.  The changes are as follows:

- Remove `tribol::integer` in favor of `int`, or other C++ fundamental types as dictated by the specific variable (e.g. `unsigned short` might make more sense for a variable that represents spatial dimension)
- Define `tribol::IndexT` as `axom::IndexType` for indexing into axom types (e.g. `axom::Array`).  Note the use of `tribol::IndexT` still needs to be checked in source files.  This will be done in future PRs.
- Rename `tribol::real` as `tribol::RealT` and ensure proper usage throughout source files.  `tribol::RealT` is defined as a `double`; if the user attempts to map it to `float` by defining `TRIBOL_USE_SINGLE_PRECISION`, an error will occur since Tribol has not been tested with single precision.
- Rename `tribol::containerArray` as `tribol::ArrayT` and define it as an `axom::Array` (with template arguments) instead of `std::vector`
- Rename `tribol::CommType` to `tribol::CommT` to be consistent with other Tribol type names

Note, this PR does not address types in redecomp or in the Fortran/C interface.

### DIFF
--- a/src/examples/common_plane.cpp
+++ b/src/examples/common_plane.cpp
@@ -19,6 +19,11 @@
 #include "tribol/physics/CommonPlane.hpp"
 #include "tribol/geom/GeomUtilities.hpp"
 
+#ifdef TRIBOL_USE_UMPIRE
+// Umpire includes
+#include "umpire/ResourceManager.hpp"
+#endif
+
 #include "axom/slic.hpp"
 
 // C/C++ includes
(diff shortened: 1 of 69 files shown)


ebchin:
Rolled into #86

PR #77, opened 2024-04-13
Unified singleton data manager class

ebchin:
This PR combines the `CouplingSchemeManager` and `MeshManager` into a templated `DataManager` class.  The underlying mesh and coupling scheme data are stored in a `std::unordered_map<IndexT, T>`, where `T = CouplingScheme` or `MeshData`.  Changes are made in source to ensure the keys sent to the `unordered_map` are of type `IndexT`, and consistency in naming the key in the source is introduced.  `mesh_id` or `mesh_id#` are used for `MeshManager` keys and `cs_id` is used for `CouplingSchemeManager` keys.


ebchin:
Rolled into #86

PR #78, opened 2024-04-30
Update spack commit to required commit

white238:
This was missed in my previous PR #48


srwopschall (approved):
@white238 Thanks for this. I got it working. Do we want to update uberenv somewhere?

ebchin (approved):
Thanks @white238!

PR #79, opened 2024-05-01
Guard PolyReorder call for overlaps with more than 2 vertices

srwopschall:
Testing showed overlaps that were degenerating (due to a bug or entirely plausible due to edge length tolerancing) to 2 or less vertices were triggering an error in a vertex reordering routine. To solve this:
- Guard the call to the vertex ordering routine and a few other lines of code with conditional for non-degenerate overlaps
- It is ok to have a degenerate overlap in many cases so this should never have triggered an error
- Use a debug log statement in the vertex ordering routine in case it is every called with <3 vertices in the polygon.

### DIFF
--- a/src/tribol/geom/GeomUtilities.cpp
+++ b/src/tribol/geom/GeomUtilities.cpp
@@ -835,15 +835,24 @@ FaceGeomError Intersection2DPolygon( const real* const RESTRICT xA,
       PolyReorder( polyXTemp, polyYTemp, numPolyVert );
 
       // check length of segs against tolerance and collapse short segments if necessary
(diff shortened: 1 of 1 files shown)


ebchin (approved):

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1237,6 +1249,14 @@ FaceGeomError CheckPolySegs( const real* const RESTRICT x, const real* const RES
       }
    }
 
+   // check to make sure numNewPoints > 3 for valid overlap polygons prior
```
ebchin: numNewPoints >= 3

PR #81, opened 2024-05-02
Redecomp Auto-Contact

srwopschall:
Fixed the mfem mesh data to allow for auto contact when constructing connectivity, etc.


ebchin (approved):
thanks for getting this up @srwopschall!

PR #82, opened 2024-05-09
Fix serial build

liu15:


ebchin (approved):
Thanks @liu15 !  I'll add a serial CI build so we can catch things like this (issue added).

PR #84, opened 2024-05-15
Fix Gap Used in Common-Plane Timestep Vote

srwopschall:
- Refactors the gap used in the common plane timestep calculation to reset the existing gap to g=0
- This fix avoids timestep crashes when the existing gap, or the velocity projected gap is slightly beyond the timestep interpenetration amount (as a factor of the element thickness).
- This fix also allows for a soft contact response without a timestep crash in the presence of larger amounts of interpenetration.
- Tests and comments have been updated

### DIFF
--- a/src/tribol/mesh/CouplingScheme.cpp
+++ b/src/tribol/mesh/CouplingScheme.cpp
@@ -1443,7 +1443,7 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
 
       real dt1 = 1.e6;  // initialize as large number
       real dt2 = 1.e6;  // initialize as large number
-      real alpha = 1.0; // multiplier on timestep estimate
+      real alpha = parameters.timestep_scale; // multiplier on timestep estimate
       bool dt1_check1 = false;
       bool dt2_check1 = false;
       bool dt1_vel_check = false;
--- a/src/tribol/mesh/CouplingScheme.cpp
+++ b/src/tribol/mesh/CouplingScheme.cpp
@@ -1532,12 +1532,13 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
 
       } // end case 1
 
-      ///////////////////////////////////////////////////////////
-      // 2. Velocity projection exceeds interpen tolerance     // 
-      //    Note: This is performed for all contact candidates //
-      //          even if they are not 'in contact' per the    //
-      //          common-plane method                          //
-      ///////////////////////////////////////////////////////////
+      ////////////////////////////////////////////////////////////////////////
+      // 2. Velocity projection exceeds max interpenetration                // 
+      //                                                                    // 
+      //    Note: This is performed for all contact candidates even if they //
+      //          are not 'in contact' per the common-plane method. Every   //
+      //          contact candidate has a contact plane                     //
+      ////////////////////////////////////////////////////////////////////////
 
       {
          // compute delta between velocity projection of face-projected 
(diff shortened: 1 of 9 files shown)


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1482,46 +1467,50 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
...
+         // This calculation RESETS the current gap to be g = 0, and computes a timestep
+         // such that the velocity projection of the overlap-to-face projected overlap 
+         // centroid does not exceed the max allowable gap. 
+         //
+         // This avoid a timestep crash in the case that the current gap barely exceeds 
+         // the max allowable and also allows a soft contact response with interpen
+         // in excess of the max allowable gap without causing timestep crashes.
+         //
+         // v1_dot_n1 > 0 and v2_dot_n2 > 0 for further interpen
+         dt1 = (dt1_check1) ? alpha * max_delta1 / v1_dot_n1 : dt1;
```
liu15, line 1511: Can we expose alpha to the host code?  This will allow me to test the timestep vote in our host code in the way that we would actually use it.

srwopschall: Added. Default is 1.0. Alpha defaults to 1.0 if user specifies alpha <= 0. Let's chat about how you would use this in your host code so I can apply this logic to mine.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1274,52 +1274,47 @@ void CouplingScheme::computeTimeStep(real &dt)
...
    ContactPlaneManager& cpMgr = ContactPlaneManager::getInstance();
-   //int num_sides = 2; // always 2 sides in a single coupling scheme
    int dim = this->spatialDimension();
    int numNodesPerCell1 = mesh1.m_numNodesPerCell;
    int numNodesPerCell2 = mesh2.m_numNodesPerCell;
 
-   // loop over each interface pair. Even if pair is not in contact, 
-   // we still do a velocity projection for that proximate face-pair 
-   // to see if interpenetration next cycle 'may' be too much
+   // Loop over interface pairs and check exising gaps for pairs in
```
liu15: spelling: existing

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1482,46 +1467,50 @@ void CouplingScheme::computeCommonPlaneTimeStep(real &dt)
...
+         //
+         // This avoid a timestep crash in the case that the current gap barely exceeds 
+         // the max allowable and also allows a soft contact response with interpen
+         // in excess of the max allowable gap without causing timestep crashes.
+         //
+         // v1_dot_n1 > 0 and v2_dot_n2 > 0 for further interpen
+         dt1 = (dt1_check1) ? alpha * max_delta1 / v1_dot_n1 : dt1;
+         dt2 = (dt2_check1) ? alpha * max_delta2 / v2_dot_n2 : dt2;
+
+         //std::cout << "dt1_check1, delta1 and v1_dot_n1: " << dt1_check1 << ", " << max_delta1 << ", " << v1_dot_n1 << std::endl;
```
liu15, line 1515: do we want to remove this commented code or have a way of enabling it?

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Fixed.

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Since the timestep vote is in development or experimental pending future testing, let's keep it in here. As I went through this I was thinking of adding a `timestep_debug` flag where data like this could be written to a file. @liu15 Thoughts? I can add that in this PR or punt for now.

srwopschall: @liu15 - Probably easy enough to add that rather on punting on a small but useful thing we just spend the time to add it?

liu15: I'm ok with it as is.  Up to you.

liu15 (approved):

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Ok. I'll leave it but make a note to keep the print statements.

PR #89, opened 2024-07-10
Redecomp: Improved entity out of balance calculation

ebchin:
Entity out of balance tolerance is now computed given the total number of entities the RCB must split on each node of the binary tree.  Better output for when the RCB decomposition fails.  Default behavior is now a warning instead of an error.


src/redecomp/partition/RCB.cpp
```diff
@@ -219,11 +223,14 @@ BisecTree<RCBInfo<NDIMS>> RCB<NDIMS>::BuildProblemTree(
 
         if (axis_ok) break;
       }
-      // none of the axes worked.  terminate the program
-      if (!axis_ok)
-      {
-        SLIC_ERROR("RCB domain decomposition unsuccessful.");
-      }
+      // none of the axes worked.  issue a warning but continue.
+      SLIC_WARNING_ROOT_IF(!axis_ok,
```
srwopschall, line 227: Per recent testing, this was still triggered quite a bit and overloaded the simulation output. Can we downgrade to a debug print and on root (if possible) until we have a better handle on what might be going on?

src/redecomp/partition/RCB.cpp
```diff
@@ -219,11 +223,14 @@ BisecTree<RCBInfo<NDIMS>> RCB<NDIMS>::BuildProblemTree(
 
         if (axis_ok) break;
       }
-      // none of the axes worked.  terminate the program
-      if (!axis_ok)
-      {
-        SLIC_ERROR("RCB domain decomposition unsuccessful.");
-      }
+      // none of the axes worked.  issue a warning but continue.
+      SLIC_DEBUG_ROOT_IF(!axis_ok,
```
srwopschall: @ebchin - Can we do this to suppress this message during normal (i.e. log level = warning) runs in some host codes? Does this need to be further tested? Should we create an issue?

srwopschall (approved):
@ebchin - Approving to get this and associated changes in.

src/redecomp/partition/RCB.cpp
srwopschall: Will patch as appropriate in host codes for now. Issue created.

src/redecomp/partition/RCB.cpp
srwopschall: Will patch in host codes for now. This should be a warning and further investigated.

PR #90, opened 2024-07-17
Modify Contact Case to Include TIED

srwopschall:
Tied contact isn't really a contact model, it is a specific 'case' in which contact enforcement is used. The contact model is meant to handle constitutive type models/response. Here we move TIED to the contact case, and delineate tied in the normal direction from tied 'full', which includes normal and tangential (there is no foreseeable use case for tied only in the tangential direction).

- Move TIED to be a contact case
- Include TIED_NORMAL (implemented) and TIED_FULL (tangential component not yet implemented).
- Pushed these changes down to the physics.
- Light refactor of the switch statements used to call the common-plane routines (more invasive design refactor beyond the scope of this PR)
- Added plumbing for adhesion based separation law in contact model, though not yet implemented.
- Updated all coupling scheme checks accordingly (I think :) )


src/tribol/common/Parameters.hpp
```diff
@@ -97,6 +97,8 @@ enum ContactCase
 {
```
ebchin, line 104: It's hard for me to remember what's a contact case vs. a model vs. a mode vs. a method.  Could you add a short definition of what each of these is?

src/tribol/common/Parameters.hpp
```diff
@@ -118,11 +120,11 @@ enum ContactMethod // all mortar methods go first
  */
 enum ContactModel
```
ebchin, line 135: Is this to define the tangential contact response?  What's the difference between `NO_CONTACT` and `FRICTIONLESS`?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -67,14 +67,14 @@ void registerMfemCouplingScheme( integer cs_id,
       {
          pressure_vdim = 1;
       }
-      else if (contact_model == TIED || contact_model == COULOMB)
+      else if (contact_model == TIED_NORMAL || contact_model == COULOMB)
```
ebchin: This should probably be `TIED_FULL` and `pressure_vdim = 1` for `TIED_NORMAL`

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -67,14 +67,14 @@ void registerMfemCouplingScheme( integer cs_id,
...
       {
          pressure_vdim = mesh.SpaceDimension();
       }
       else
       {
          SLIC_ERROR_ROOT("Unsupported contact model. "
-           "Only FRICTIONLESS, TIED, and COULOMB supported.");
+           "Only FRICTIONLESS, TIED_NORMAL, and COULOMB supported.");
```
ebchin:
```suggestion
           "Only FRICTIONLESS, TIED_FULL, TIED_NORMAL, and COULOMB supported.");
```

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -517,29 +517,39 @@ bool CouplingScheme::isValidCase()
       this->m_contactCase = NO_CASE;
    }
 
-   // specify auto-contact specific interpenetration check and verify 
-   // element thicknesses have been registered
    parameters_t& params = parameters_t::getInstance();
-   if (this->m_contactCase == AUTO)
-   { 
-      params.auto_interpen_check = true;
+   if (this->m_contactMethod == COMMON_PLANE)
```
ebchin, line 525: What does `ContactCase::AUTO` do after this change?

src/tribol/common/Parameters.hpp
srwopschall: Added. This adds clarification for sure.

src/tribol/common/Parameters.hpp
srwopschall: The contact model is used to specify some constitutive response across an interface. For now, this will generally be in the tangential direction. That said, if for some reason a host code wants to lose contact in a coupling scheme, this would be the place to specify no contact, or no interaction/physics. There is not a use case for this right now, and there may never be one, but for now I'll keep this. It would ease a host-code integration where they can keep the coupling scheme and all tribol calls (e.g. not guarded), but switch to no contact for whatever reason.

src/tribol/interface/mfem_tribol.cpp
srwopschall: The only issue here is that this part of the code is treating model and case options (the tied should be moved to check contact_case) that aren't supported or implemented. This conflates the coupling scheme checks with additional checks here that assign the pressure_vdim, exposing the potential for some inconsistency. On the other hand, if here we treat the whole kitchen sink, Tribol shouldn't get passed coupling scheme apply for unsupported/umimplemented combinations. With that in mind, perhaps I expand this on the case and model to include all enumerated options just for completeness?

src/tribol/mesh/CouplingScheme.cpp
srwopschall: I'll test elsewhere, but in principle there is no change. What this design does preclude is specifying auto contact with a tied use case. I would argue that auto, however, is for instances when you truly do not know what is in contact with what, or the number of contact surfaces in a coupling scheme is too burdensome to individually specify. In a TIED application, you know exactly what is in contact and specifying surface boundary attributes shouldn't be a barrier to entry. Physically speaking, too, auto will most likely treat multiple mesh interfaces, especially self contact and cases where you don't know what will be in contact, but TIED specifically treats different topological entities where specifying separate boundary attributes should be easy enough to do.

src/tribol/interface/mfem_tribol.cpp
srwopschall: See comment above about the else if statement.

src/tribol/common/Parameters.hpp
```diff
@@ -74,12 +74,10 @@ enum VisType
 
 /*!
  * \brief Tribol supports the following different contact modes:
```
ebchin: Might need to update this blurb with the new definition below

src/tribol/interface/mfem_tribol.cpp
ebchin: `TIED_NORMAL` only treats pressure in the normal direction (1 dimension) while `TIED_FULL` has a traction vector (in all spatial dimensions)?  It probably doesn't matter for now, since `pressure_vdim != 1` isn't used anywhere.  Maybe just set `pressure_vdim = 1` and add a note/TODO that this will need to be updated when we have more than a normal response.

More generally, we'll have to think how we want to handle/store this with tangential response and with full tied contact.  For tied contact, the components will likely align with the Cartesian axes, but potentially not for normal + tangential response.

src/tribol/mesh/CouplingScheme.cpp
ebchin: Makes sense.  We might want to document this, perhaps in `Parameters.hpp`?

ebchin (approved):
Thanks @srwopschall.  A few more minor comments, but I went ahead and approved.

src/tribol/interface/mfem_tribol.cpp
srwopschall: That is right. The normal component is only stored as a pressure, but the normal traction would simply be dotting that pressure with the common plane normal. The tangential component would likely be stored as a full traction. All of that is internal. These would still be integrated to give nodal force vectors, which is what is passed through the API.

I'll simplify this part of the code and put the TODO.

srwopschall: Looking at this again, these are checks to set the pressure dim inside an if-statement when using Lagrange multipliers. Does this occur prior to calling coupling scheme init()? If so, we just check for Lagrange multipliers here, and not whether that enforcement is combined with an allowed method etc. May not be a huge deal, but looks like we create some operators here that would still happen even if the coupling scheme is not valid. E.g. later when init() is called, we may find that the coupling scheme is not valid (e.g. Lagrange multipliers with common plane (not allowed) or with coulomb friction (not implemented).

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Added to the doxygen comment for the ContactCase.

src/tribol/common/Parameters.hpp
srwopschall: Updated.

src/tribol/interface/mfem_tribol.cpp
ebchin: That's potentially true, but there isn't a way to create the pressure grid function after calling `CouplingScheme::init()`.  `CouplingScheme::init()` is trigged by calling `update()`, and when `update()` is called, all of the MFEM and Tribol data structures have already been created.  I think we'll have to leave this as is, pending a pretty major refactor.

ebchin: I think, in an ideal world, the MFEM API would not allow Tribol to get into an invalid state (i.e. `CouplingScheme::init()` should never fail), but I don't think we're there yet :smiley:

PR #94, opened 2024-08-05
Reset null mesh value with each CouplingScheme validity check

ebchin:
@zatkins-work found an issue with Tribol not resetting the null meshes each time a `CouplingScheme` is checked for validity.  Currently, the variable `CouplingScheme::m_nullMeshes` is true if any of the meshes have ever been null.  After the fix, the value of `CouplingScheme::m_nullMeshes` will only be true if either of the meshes is currently null.

Thanks for the fix @zatkins-work!

### DIFF
--- a/src/tribol/mesh/CouplingScheme.cpp
+++ b/src/tribol/mesh/CouplingScheme.cpp
@@ -389,6 +389,9 @@ bool CouplingScheme::isValidCouplingScheme()
    MeshData & mesh1 = meshManager.GetMeshInstance( this->m_meshId1 );
    MeshData & mesh2 = meshManager.GetMeshInstance( this->m_meshId2 );
 
+   // set boolean for null meshes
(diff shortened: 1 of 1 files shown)


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -403,13 +406,6 @@ bool CouplingScheme::isValidCouplingScheme()
    {
       return false;
    }
-   
-   // set boolean for null meshes
-   if ( mesh1.m_numCells <= 0 || mesh2.m_numCells <= 0 )
-   {
-      this->m_nullMeshes = true;
-      valid = true; // a null-mesh coupling scheme should still be valid
```
srwopschall, line 411: Just to be thorough, I see that we no longer need to set valid = true because valid is initialized to true, and stays true when we set the boolean for null meshes in the new lines of code.

srwopschall (approved):
Looks good.

PR #103, opened 2024-09-06
Fixed Bug in Adjancent Face Check in Geom Filter

srwopschall:
- Adjacent faces were precluded from contact ONLY for matching mesh ids, which presents issues at sharp edges or corners where the configuration and opposing normals make the faces appear to be in contact.
- This removes the conditional on the mesh ids, which doesn't mean anything anymore, and simply precludes contact between adjacent faces. This removes adjacent faces at corners from being included in contact (which is wrong). 
- The consequence is that two adjacent faces can't be in self contact, but self-contact should be resolved beyond such a local face-pair anyway.

### DIFF
--- a/src/tribol/common/Parameters.hpp
+++ b/src/tribol/common/Parameters.hpp
@@ -517,7 +517,7 @@ struct parameters_t
   bool enable_timestep_vote;     ///! True if host-code desires the timestep vote to be calculated and returned
 
   double auto_contact_len_scale_factor; ///! Sacle factor applied to element thickness for auto contact length scale
-  bool auto_contact_check;              ///! True for auto contact enabled checks (e.g. geometry specific checks etc.)
+  bool auto_contact_check;              ///! True if auto-contact checks should be enabled
 
 private:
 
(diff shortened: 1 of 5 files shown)


src/tribol/common/Parameters.hpp
```diff
@@ -517,7 +517,7 @@ struct parameters_t
   bool enable_timestep_vote;     ///! True if host-code desires the timestep vote to be calculated and returned
 
   double auto_contact_len_scale_factor; ///! Sacle factor applied to element thickness for auto contact length scale
-  bool auto_interpen_check;             ///! True if the auto-contact interpenetration check is used for full-overlap pairs
+  bool auto_contact_check;              ///! True for auto contact enabled checks (e.g. geometry specific checks etc.)
```
ebchin: "True if auto contact checks should be enabled"?

ebchin (approved):
Looks good @srwopschall

src/tribol/common/Parameters.hpp
srwopschall: Yes. Clearer.Fixed.

PR #114, opened 2024-11-19
2d/3d Bugfix and Cleanup

srwopschall:
- Cleaned up old comments related to flipping of segments between registerMesh() call and physics calculations; this is no longer an issue
- Fixed basis function evaluation at common plane integration point to use projected faces onto common plane overlap
- Fixed 2D computational geometry bug where the segment overlap calculation was comparing vector magnitudes to non-projected face areas leading to false positives.
- Introduced computational geometry test where the configuration used previously tripped a false positive overlap.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -65,6 +65,10 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 - Allow null velocity and response pointers for various use cases
 - Tolerancing bug that produced negative timestep estimates in the presence of numerically
   zero face velocities.
+- Fixed computational geometry bug in 2D/3D common plane that was using face normals in projections
+  instead of common plane normal.
+- Fixed bug in 2D segment-segment overlap calculation on common plane. The current configuration edge areas
+  were being used instead of the projected edges leading to false positives.
 
 ## [Version 0.1.0] - Release date 2023-04-21
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -67,7 +67,7 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
   zero face velocities.
 - Fixed computational geometry bug in 2D/3D common plane that was using face normals in projections
   instead of common plane normal.
-- Fixed bug in 2D segment-segment overlap calculation on common plane. The current configuration edge areas
+- Fixed bug in 2D segment-segment overlap calculation on common plane. The current configuration edge lengths
   were being used instead of the projected edges leading to false positives.
 
 ## [Version 0.1.0] - Release date 2023-04-21
--- a/src/tests/tribol_comp_geom.cpp
+++ b/src/tests/tribol_comp_geom.cpp
@@ -292,6 +292,88 @@ TEST_F( CompGeomTest, poly_area_centroid_2 )
    EXPECT_LE( diff_mag, tol );
 }
 
+TEST_F( CompGeomTest, should_produce_no_overlap )
+{
+   // this is a configuration from testing that is/was producing an overlap for
+   // non-overlapping edges, which in turn produced negative basis function
+   // evaluations
+   constexpr int dim = 2;
+   constexpr int numVerts = 2;
+   RealT xy1[dim*numVerts];
+   RealT xy2[dim*numVerts];
+
+   //   Edge 1 (x,y)         Edge 2 (x,y)         Proj Edge 1 (x,y)   Proj Edge 2 (x,y)     Overlap (x,y)        Integration Point
+   //0.324552, 0.625596   4.59227e-17, 0.752178  0.336421, 0.657206  0.0103139, 0.779648   0.162068, 0.722669   0.161894, 0.722735
+   //0.16206, 0.722646    0.161705,    0.72276   0.162068, 0.722669  0.16172,   0.722800   0.16172,  0.722800
+
+   // this geometry should NOT be in contact
+   xy1[0] = 0.324552;
+   xy1[1] = 0.625596;
+   xy1[2] = 0.16206;
+   xy1[3] = 0.722646;
+
+   xy2[0] = 4.59227e-17;
+   xy2[1] = 0.752178;
+   xy2[2] = 0.161705;
+   xy2[3] = 0.72276;
+
+   RealT x1[numVerts];
+   RealT y1[numVerts];
+   RealT x2[numVerts];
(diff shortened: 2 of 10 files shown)


RELEASE-NOTES.md
```diff
@@ -65,6 +65,10 @@ Changelog](http://keepachangelog.com/en/1.0.0/).
 - Allow null velocity and response pointers for various use cases
 - Tolerancing bug that produced negative timestep estimates in the presence of numerically
   zero face velocities.
+- Fixed computational geometry bug in 2D/3D common plane that was using face normals in projections
+  instead of common plane normal.
+- Fixed bug in 2D segment-segment overlap calculation on common plane. The current configuration edge areas
```
ebchin:
```suggestion
- Fixed bug in 2D segment-segment overlap calculation on common plane. The current configuration edge lengths
```

src/tests/tribol_comp_geom.cpp
```diff
@@ -292,36 +292,321 @@ TEST_F( CompGeomTest, poly_area_centroid_2 )
...
+}
+
+TEST_F( CompGeomTest, coincident_vertices_full_overlap )
+{
+   constexpr int dim = 2;
+   constexpr int numVerts = 2;
+   RealT xy1[dim*numVerts];
+   RealT xy2[dim*numVerts];
+
+   // this geometry should NOT be in contact
```
ebchin: Can you give a brief explanation about why tribol shouldn't think these edges aren't in contact?  Maybe something like "edges are too far away so xxx check will fail."

src/tests/tribol_comp_geom.cpp
```diff
@@ -292,36 +292,321 @@ TEST_F( CompGeomTest, poly_area_centroid_2 )
...
+
+   EXPECT_EQ( update_err, 0 );
+
+   tribol::CouplingSchemeManager& couplingSchemeManager = 
+         tribol::CouplingSchemeManager::getInstance();
+
+   tribol::CouplingScheme* couplingScheme = 
+      &couplingSchemeManager.at( 0 );
+
+   EXPECT_EQ( 1, couplingScheme->getNumActivePairs() );
```
ebchin, line 446: Ahh, if these should be in contact, the comment above should be updated.

src/tests/tribol_comp_geom.cpp
```diff
@@ -292,36 +292,321 @@ TEST_F( CompGeomTest, poly_area_centroid_2 )
...
+}
+
+TEST_F( CompGeomTest, coincident_vertex_no_overlap )
+{
+   constexpr int dim = 2;
+   constexpr int numVerts = 2;
+   RealT xy1[dim*numVerts];
+   RealT xy2[dim*numVerts];
+
+   // this geometry should NOT be in contact
```
ebchin: A bigger description might be helpful here too.  Just looking at the coords, there should be an overlap of 1e-8.  What tolerances etc. control whether this is in contact?

src/tests/tribol_comp_geom.cpp
```diff
@@ -292,36 +292,321 @@ TEST_F( CompGeomTest, poly_area_centroid_2 )
...
+   constexpr int numVerts = 2;
+   RealT xy1[dim*numVerts];
+   RealT xy2[dim*numVerts];
+
+   xy1[0] = 1.0;
+   xy1[1] = 0.0;
+   xy1[2] = 0.0;
+   xy1[3] = 0.0;
+
+   xy2[0] = -1.;
```
ebchin, line 547: Which tolerance controls this now being considered an overlap?  A brief comment would be helpful.

src/tribol/integ/Integration.cpp
```diff
@@ -37,65 +38,98 @@ TRIBOL_HOST_DEVICE void EvalWeakFormIntegral< COMMON_PLANE, SINGLE_POINT >
...
-   RealT cxf2[3] = { 0., 0., 0. }; // vertex avg. centroid of face 2
-
-   // compute vertex averaged centroid of each face for the point-normal data
-   VertexAvgCentroid( elem.faceCoords1, elem.dim, elem.numFaceVert,
-                      cxf1[0], cxf1[1], cxf1[2] );
-   VertexAvgCentroid( elem.faceCoords2, elem.dim, elem.numFaceVert,
-                      cxf2[0], cxf2[1], cxf2[2] );
-
-   // project the overlap centroid to each face
+   // debug: leave commented out so we don't enter loop
```
ebchin, line 41: If we had an `#ifdef TRIBOL_DEBUG`, this would be a good place to use it.

src/tribol/integ/Integration.cpp
```diff
@@ -26,7 +26,8 @@ TRIBOL_HOST_DEVICE void EvalWeakFormIntegral< COMMON_PLANE, SINGLE_POINT >
                            RealT * const integ2 )
 {
    // compute the area centroid of the overlap polygon,
-   // which serves as the single integration point
+   // or vertex avg. centroid of the overlap segment, which 
+   // serves as the single integration point
```
ebchin, line 30: Thanks for adding this

ebchin (approved):
Thanks for diving in and finding these bugs @srwopschall!  A few minor comments, but otherwise looks good

RELEASE-NOTES.md
srwopschall: Fixed.

src/tests/tribol_comp_geom.cpp
srwopschall: I updated the comments about the nature of the face-pair configuration for all new tests.

src/tests/tribol_comp_geom.cpp
srwopschall: Fixed.

src/tests/tribol_comp_geom.cpp
srwopschall: Added a comment giving more clarity.

src/tests/tribol_comp_geom.cpp
srwopschall: Added a comment indicating that the contact area frac set in setContactAreaFrac() controls the tolerancing or area thresholding.

src/tribol/integ/Integration.cpp
srwopschall: Definitely. I think we can address this when we do our next overhaul of logging such that debug loops are properly guarded instead of commented out.

PR #115, opened 2024-12-02
add macos spack config

ebchin:
Adds a spack config for Mac OS Sequoia and fixes a couple of small errors caught by the compiler.


srwopschall (approved):
Thanks @ebchin

PR #122, opened 2025-01-10
Timestep Velocity Projection Modification

srwopschall:
The current velocity projection check performs a velocity projection of the overlap centroid as projected to one face in a face-pair and compares this against the face-projected overlap centroid on the OTHER face in the CURRENT configuration. If both faces are moving toward one another, this underestimates the amount of interpenetration. 

This PR introduces the change where the velocity projection of the overlap centroid as projected to one face is compared against the _velocity projected_ overlap centroid (on the other face), which represents a PROJECTED configuration rather than the current configuration. This will more accurately catch excessive interpenetration for nonzero incoming velocities and trigger a contact timestep vote.

### DIFF
--- a/src/tribol/mesh/CouplingScheme.cpp
+++ b/src/tribol/mesh/CouplingScheme.cpp
@@ -1514,16 +1514,22 @@ void CouplingScheme::computeCommonPlaneTimeStep( RealT& dt )
                 ////////////////////////////////////////////////////////////////////////
 
                 {
-                  // compute delta between velocity projection of face-projected
-                  // overlap centroid and the OTHER face's face-projected overlap
-                  // centroid
-                  RealT proj_delta_x1 = plane.m_cXf1 + dt * vel_f1[0] - plane.m_cXf2;
-                  RealT proj_delta_y1 = plane.m_cYf1 + dt * vel_f1[1] - plane.m_cYf2;
+                  // compute delta between the velocity projection of each face-projected
+                  // overlap centroid. First project each point.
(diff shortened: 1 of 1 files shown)


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1514,17 +1514,23 @@ void CouplingScheme::computeCommonPlaneTimeStep( RealT& dt )
...
-                  RealT proj_delta_y2 = plane.m_cYf2 + dt * vel_f2[1] - plane.m_cYf1;
+                  RealT proj_delta_x2 = plane.m_cXf2 + dt * vel_f2[0];
+                  RealT proj_delta_y2 = plane.m_cYf2 + dt * vel_f2[1];
                   RealT proj_delta_z2 = 0.;
 
+                  // subtract off the other face's velocity projected point
+                  proj_delta_x1 -= proj_delta_x2;
+                  proj_delta_y1 -= proj_delta_y2;
+
+                  proj_delta_x2 -= proj_delta_x1;
```
ebchin: Is this right?  I think we want to `-= proj_delta_x1` before modifying it on the above lines (L1528)?  Same with y and z below.

ebchin (changes requested):

ebchin (approved):

src/tribol/mesh/CouplingScheme.cpp
srwopschall: Good catch. I think I fixed this by using the original `proj_delta_x1` based on the velocity projection of that point, not the modified one as is done here (modified on L1528 and used on L1531).

PR #123, opened 2025-01-23
enable setting number of redecomp ranks

ebchin:
This PR makes two small changes

- Allows the user to specify how many ranks the redecomp mesh should be created on
- Targets about 10 element pairs/rank as a new minimum default

### DIFF
--- a/src/redecomp/RedecompMesh.cpp
+++ b/src/redecomp/RedecompMesh.cpp
@@ -15,12 +15,12 @@
 
 namespace redecomp {
 
-RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, PartitionType method, int num_ranks )
-    : RedecompMesh( parent, DefaultGhostLength( parent ), method, num_ranks )
+RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, PartitionType method, int n_ranks )
+    : RedecompMesh( parent, DefaultGhostLength( parent ), method, n_ranks )
 {
 }
 
-RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length, PartitionType method, int num_ranks )
+RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length, PartitionType method, int n_ranks )
     : parent_{ parent }, mpi_{ parent.GetComm() }
 {
   // build partitioner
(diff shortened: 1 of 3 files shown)


src/redecomp/RedecompMesh.cpp
```diff
@@ -54,20 +54,26 @@ RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length, Pa
...
 
-  // preclude degenerate case where num elements/2 < num ranks
-  // factor of 2 on num elements are due to elements being paired for contact
-  auto n_parts = std::min( parent.GetNRanks(), ( static_cast<int>( parent.GetGlobalNE() ) + 1 ) / 2 );
-  p2r_elems_ = BuildP2RElementList( *partitioner, n_parts, ghost_length );
+  auto parent_n_els = static_cast<int>( parent.GetGlobalNE() );
+  if ( n_ranks == 0 ) {
+    // preclude degenerate case where num elements/2 < num ranks
+    // factor of 2 on num elements are due to elements being paired for contact
+    // additional factor of 10 targets about 10 pairs/rank
```
srwopschall: Mention for clarification that this logic assumes conforming mesh, but that this is used as a lower limit. A nonconforming mesh would result in more than a minimum of 10 pairs per rank (or something like that).

src/redecomp/RedecompMesh.cpp
```diff
@@ -77,27 +83,31 @@ RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length,
...
   }
 
-  // preclude degenerate case where num elements < num ranks
-  auto n_parts = std::min( parent.GetNRanks(), static_cast<int>( parent.GetGlobalNE() ) );
+  auto parent_n_els = static_cast<int>( parent.GetGlobalNE() );
+  if ( n_ranks == 0 ) {
+    // preclude degenerate case where num elements/2 < num ranks
+    // factor of 2 on num elements are due to elements being paired for contact
+    // additional factor of 10 targets about 10 pairs/rank
+    n_ranks = std::max( std::min( parent.GetNRanks(), ( parent_n_els + 1 ) / 2 / 10 ), 1 );
```
srwopschall, line 114: see note above about conforming mesh as minimum case.

src/redecomp/RedecompMesh.cpp
```diff
@@ -54,20 +54,26 @@ RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length, Pa
       SLIC_ERROR_ROOT( "Only 2D and 3D meshes are supported." );
   }
 
-  // preclude degenerate case where num elements/2 < num ranks
-  // factor of 2 on num elements are due to elements being paired for contact
-  auto n_parts = std::min( parent.GetNRanks(), ( static_cast<int>( parent.GetGlobalNE() ) + 1 ) / 2 );
-  p2r_elems_ = BuildP2RElementList( *partitioner, n_parts, ghost_length );
+  auto parent_n_els = static_cast<int>( parent.GetGlobalNE() );
+  if ( n_ranks == 0 ) {
```
srwopschall, line 61: Can you clarify why this check is for n_ranks == 0? Is this just saying that n_ranks is initialized to 0 and then needs to be set, but if != 0 that means we have already set it appropriate according to this new criterion?

src/redecomp/RedecompMesh.cpp
```diff
@@ -77,27 +83,31 @@ RedecompMesh::RedecompMesh( const mfem::ParMesh& parent, double ghost_length,
...
       break;
     }
     default:
       SLIC_ERROR_ROOT( "Only 2D and 3D meshes are supported." );
   }
 
-  // preclude degenerate case where num elements < num ranks
-  auto n_parts = std::min( parent.GetNRanks(), static_cast<int>( parent.GetGlobalNE() ) );
+  auto parent_n_els = static_cast<int>( parent.GetGlobalNE() );
+  if ( n_ranks == 0 ) {
```
srwopschall, line 109: It is hard for me to tell in github but this if-block and rank calculation appears twice. Why is that needed twice?

srwopschall (approved):
@ebchin - Thanks for this PR. I have a few comments for clarification, but looks good.

src/redecomp/RedecompMesh.cpp
ebchin: This could probably be cleaned up, but there are currently two constructors that use this logic.

src/redecomp/RedecompMesh.cpp
ebchin: See doxygen comment for what happens when n_ranks == 0.  I added some checks for preventing n_ranks < 0 and n_ranks > number of MPI ranks.

PR #125, opened 2025-01-24
Add proximity scale for binning

ebchin:
Adds a proximity scaling for binning.  Sets the default at 4.0 with a minimum of 2.0 (enforced through the API call `setBinningProximityScale()`).

### DIFF
--- a/src/tribol/common/Parameters.hpp
+++ b/src/tribol/common/Parameters.hpp
@@ -463,13 +463,14 @@ struct EnforcementOptions {
 struct Parameters {
   CommT problem_comm = TRIBOL_COMM_WORLD;  ///! MPI communicator for the problem
 
-  RealT binning_proximity_scale = 4.0;  ///! Approximate element length multiplier for including pairs in coarse binning
-  RealT overlap_area_frac = 1.0e-8;     ///! Ratio of overlap area to largest face area for contact inclusion
-  RealT gap_tol_ratio = 1.0e-12;        ///! Ratio for determining tolerance for active contact gaps
-  RealT gap_separation_ratio = 0.75;    ///! Ratio for determining allowable separation in geometric filtering
-  RealT gap_tied_tol = 0.1;             ///! Ratio for determining max separation tied contact can support
-  RealT len_collapse_ratio = 1.0e-8;    ///! Ratio of face length providing topology collapse length tolerance
-  RealT projection_ratio = 1.0e-10;     ///! Ratio for defining nonzero projections
+  RealT binning_proximity_scale =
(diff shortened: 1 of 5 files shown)


src/tribol/common/Parameters.hpp
```diff
@@ -463,12 +463,13 @@ struct EnforcementOptions {
 struct Parameters {
   CommT problem_comm = TRIBOL_COMM_WORLD;  ///! MPI communicator for the problem
 
-  RealT overlap_area_frac = 1.0e-8;   ///! Ratio of overlap area to largest face area for contact inclusion
-  RealT gap_tol_ratio = 1.0e-12;      ///! Ratio for determining tolerance for active contact gaps
-  RealT gap_separation_ratio = 0.75;  ///! Ratio for determining allowable separation in geometric filtering
-  RealT gap_tied_tol = 0.1;           ///! Ratio for determining max separation tied contact can support
-  RealT len_collapse_ratio = 1.0e-8;  ///! Ratio of face length providing topology collapse length tolerance
-  RealT projection_ratio = 1.0e-10;   ///! Ratio for defining nonzero projections
+  RealT binning_proximity_scale = 4.0;  ///! Approximate element length multiplier for including pairs in coarse binning
```
srwopschall: Don't need to say "approximate" cause it really isn't approximate to anything. Suggest: Element length multiplier for coarse binning and proximity detection inclusion

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -410,13 +407,14 @@ class GridSearch : public SearchBase {
     const auto mesh1 = m_coupling_scheme->getMesh1().getView();
     const auto mesh2 = m_coupling_scheme->getMesh2().getView();
     auto& contactPairs = m_coupling_scheme->getInterfacePairs();
+    auto binning_proximity_scale = m_coupling_scheme->getParameters().binning_proximity_scale;
 
     // Find matches in first mesh (with index 'fromIdx')
     // with candidate elements in second mesh (with index 'toIdx')
-    int k = 0;
+    // int k = 0;  // Debug only
```
srwopschall, line 414: Just verify that this was only used for debug. Were we using this in any tests for output?

srwopschall (approved):
@chin Looks good. Minor comments.

src/tribol/search/InterfacePairFinder.cpp
ebchin: No, it was only used in commented out debug output below.  I just commented this out to silence a warning.

src/tribol/common/Parameters.hpp
ebchin: Fixed

PR #127, opened 2025-01-28
change inv iso to warning

ebchin:
Prevents a quad point that does not project in the element from crashing a sim.


srwopschall (approved):
@chin - This works, just push an error code higher up if this becomes burdensome in the output in practice

PR #128, opened 2025-02-01
Add tests and fixes for `binning_proximity_scale`

ebchin:
Adds a test for `binning_proximity_scale` to test the force response for the valid combinatorics of:
- FE order (1, 2)
- method (common plane, mortar)
- dimension (2, 3)
- binning scheme (BVH, grid, cartesian product)

Fixes bugs that were discovered along the way.

Adds a `MeshBuilder` class which simplifies generating MFEM test meshes.


src/shared/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+    mesh/MeshBuilder.cpp
+    )
+
+## setup the dependency list for tribol shared library
+set(shared_depends mfem)
+blt_list_append(TO tribol_depends ELEMENTS blt::mpi IF TRIBOL_USE_MPI )
+message(STATUS "Tribol shared library dependencies: ${shared_depends}")
+
+## create the library
+blt_add_library(
```
srwopschall, line 24: Looks like creating the shared directory as a library is the best way to treat this such that redecomp and tribol can both use it?

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -0,0 +1,162 @@
+#include "MeshBuilder.hpp"
+
+#include "axom/slic.hpp"
+#include "axom/primal.hpp"
+
+namespace shared {
+
+MeshBuilder MeshBuilder::Merged( std::initializer_list<MeshBuilder> meshes )
```
srwopschall: We talked about renaming this function. "Merge" is typically something that merges two topologically distinct geometries together. I think here you want to create a single mesh from two meshes (or something). I know we had a good name for this. Can you recall?

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+ * @brief A class for building and manipulating MFEM meshes.
+ */
+class MeshBuilder {
+ public:
+  /**
+   * @brief Creates a single mesh from a list of MeshBuilder objects.
+   * @param meshes A list of MeshBuilder objects from which the mesh will be created.
+   * @return A new MeshBuilder object representing the merged mesh.
+   */
+  static MeshBuilder Merged( std::initializer_list<MeshBuilder> meshes );
```
srwopschall: Rename for clarity

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * @brief Creates a single mesh from multiple MeshBuilder objects.
+   * @tparam Args Variadic template parameter for MeshBuilder objects.
+   * @param meshes MeshBuilder objects from which the mesh will be created.
+   * @return A new MeshBuilder object representing the merged mesh.
+   */
+  template <typename... Args>
+  static MeshBuilder Merged( Args&&... meshes );
```
srwopschall: Rename for clarity.

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+   */
+  static MeshBuilder Merged( std::initializer_list<MeshBuilder> meshes );
+
+  /**
+   * @brief Creates a single mesh from multiple MeshBuilder objects.
+   * @tparam Args Variadic template parameter for MeshBuilder objects.
+   * @param meshes MeshBuilder objects from which the mesh will be created.
+   * @return A new MeshBuilder object representing the merged mesh.
+   */
+  template <typename... Args>
```
srwopschall, line 42: Not there in my review yet, but wonder if this is used at all.

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+  static MeshBuilder Merged( Args&&... meshes );
+
+  /**
+   * @brief Creates a square mesh occupying the unit square, [0, 1]^2.
```
srwopschall, line 46: Should this take in a bounding box instead? Do we want to be constrained to using a unit square? Or, can we introduce a scaling operator?

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+   * @param n_x_els Number of elements in the x direction (>0).
+   * @param n_y_els Number of elements in the y direction (>0).
+   * @return A new MeshBuilder object representing the square mesh.
+   */
+  static MeshBuilder SquareMesh( int n_x_els, int n_y_els );
+
+  /**
+   * @brief Creates a cube mesh occupying the unit cube, [0, 1]^3.
```
srwopschall, line 54: Same question regarding unit cube.

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+   * @param n_x_els Number of elements in the x direction (>0).
+   * @param n_y_els Number of elements in the y direction (>0).
+   * @param n_z_els Number of elements in the z direction (>0).
+   * @return A new MeshBuilder object representing the cube mesh.
+   */
+  static MeshBuilder CubeMesh( int n_x_els, int n_y_els, int n_z_els );
+
+  /**
+   * @brief Creates a hypercube mesh occupying the unit hypercube, [0, 1]^dim.
```
srwopschall, line 63: Do we need this or rather, do we want to support this?

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+   * @param dim Dimension of the hypercube (between 1 and 3).
+   * @param n_els Number of elements in each direction of each dimension (>0).
```
srwopschall, line 65: If we are going to support this, do we want to specify different number of elements in each dimension? This may be useful for any given contact problem.

src/shared/mesh/MeshBuilder.hpp
```diff
@@ -0,0 +1,187 @@
...
+  static MeshBuilder HypercubeMesh( int dim, int n_els );
+
+  /**
+   * @brief Constructs a MeshBuilder object from an mfem::Mesh.
+   * @param mesh An rvalue reference to an mfem::Mesh object.
+   */
+  MeshBuilder( mfem::Mesh&& mesh );
+
+  /**
+   * @brief Translates the mesh by a given vector.
```
srwopschall, line 77: Just confirming that this translates the whole mesh, and does not scale each dimension?

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -0,0 +1,162 @@
...
+{
+  return mfem::Mesh::MakeCartesian2D( n_x_els, n_y_els, mfem::Element::QUADRILATERAL );
+}
+
+MeshBuilder MeshBuilder::CubeMesh( int n_x_els, int n_y_els, int n_z_els )
+{
+  return mfem::Mesh::MakeCartesian3D( n_x_els, n_y_els, n_z_els, mfem::Element::HEXAHEDRON );
+}
+
+MeshBuilder MeshBuilder::HypercubeMesh( int dim, int n_els )
```
srwopschall, line 30: It looks like we don't really need this HypercubeMesh() especially since Tribol just supports 2D and 3D, which here just calls MakeCartesian2D or 3D, which is done above in SquareMesh() or CubeMesh().

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+MeshBuilder&& MeshBuilder::updateAttrib( int old_attrib, int new_attrib )
+{
+  for ( int i = 0; i < mesh_.GetNE(); ++i ) {
+    if ( mesh_.GetAttribute( i ) == old_attrib ) {
+      mesh_.SetAttribute( i, new_attrib );
+    }
+  }
+  auto old_attrib_idx = mesh_.attributes.Find( old_attrib );
```
srwopschall, line 68: It sort of looks like these lines of code outside the loop over number of elements should do what `SetAttribute()` above does. Maybe not a big deal. A comment here might clarify what is being done.

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -0,0 +1,162 @@
...
+            auto vdof = mesh_.GetNodes()->FESpace()->DofToVDof( bdr_dofs[k], d );
+            vert[d] = ( *mesh_.GetNodes() )[vdof];
+          }
+          attrib_bbox.addPoint( vert );
+        }
+      }
+    }
+    if ( num_elems > 0 ) {
+      std::cout << "Boundary attribute " << current_attrib << std::endl;
+      std::cout << "  Number of elements: " << num_elems << std::endl;
```
srwopschall, line 100: Minor: "Number of boundary elements:"

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+MeshBuilder&& MeshBuilder::updateBdrAttrib( int old_attrib, int new_attrib )
+{
+  for ( int i = 0; i < mesh_.GetNBE(); ++i ) {
+    if ( mesh_.GetBdrAttribute( i ) == old_attrib ) {
+      mesh_.SetBdrAttribute( i, new_attrib );
+    }
+  }
+  auto old_attrib_idx = mesh_.bdr_attributes.Find( old_attrib );
```
srwopschall, line 117: Same comment as updateAttrib()

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // boundary element attributes of contact surface 2, the bottom surface of the top block
+    auto contact_surf_2 = std::set<int>( { 5 } );
+
+    // create a mesh with two blocks
+    // clang-format off
+    shared::ParMeshBuilder mesh( MPI_COMM_WORLD, shared::MeshBuilder::Merged( {
+      shared::MeshBuilder::SquareMesh( 1, 1 ),
+      shared::MeshBuilder::SquareMesh( 1, 1 )
+        .translate( { 0.0, 1.0 - mesh_interpenetration } )
+        .updateAttrib( 1, 2 )
```
srwopschall: Not super clear how the old attributes got assigned and where the new ones come from. Looks like surface 1 has boundary attribute 3 and surface 2 has 5. Not sure where 6 comes from, for instance. Perhaps clarify so someone could pattern match if they needed to write their own test using this test harness.

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+    r = 0.0;
+    tribol::getMfemResponse( coupling_scheme_id, r );
+    max_force_ = r.Max();
+
+    r.Print();
+
+    MPI_Allreduce( MPI_IN_PLACE, &max_force_, 1, MPI_DOUBLE, MPI_MAX, MPI_COMM_WORLD );
+  }
+
+  void SetUpCommonPlaneProblem( shared::ParMeshBuilder& mesh, const std::set<int>& contact_surf_1,
```
srwopschall: Let's rename this routine. You have SetUp2DCommonPlaneProblem() and the 3D variant, which does setup the problem, but this routine really registers the coupling scheme and associated data. Maybe something like RegsiterCommonPlaneScheme() or something along those lines.

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+    tribol::registerMfemCouplingScheme( coupling_scheme_id, mesh1_id, mesh2_id, mesh, mesh.getNodes(), contact_surf_1,
+                                        contact_surf_2, tribol::SURFACE_TO_SURFACE, tribol::NO_CASE,
+                                        tribol::COMMON_PLANE, tribol::FRICTIONLESS, tribol::PENALTY, binning_method );
+    tribol::setMfemKinematicConstantPenalty( coupling_scheme_id, penalty, penalty );
+    tribol::setBinningProximityScale( coupling_scheme_id, binning_proximity );
+
+    UpdateTribol( mesh, coupling_scheme_id );
+  }
+
+  void SetUp2DCommonPlaneProblem( tribol::BinningMethod binning_method )
```
srwopschall, line 80: Minor: Setup is usually one word so "U" would be lower case :)

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // boundary element attributes of contact surface 2, the bottom surface of the top block
+    auto contact_surf_2 = std::set<int>( { 7 } );
+
+    // create a mesh with two cubes
+    // clang-format off
+    shared::ParMeshBuilder mesh( MPI_COMM_WORLD, shared::MeshBuilder::Merged( {
+      shared::MeshBuilder::HypercubeMesh( dim, 1 ),
+      shared::MeshBuilder::HypercubeMesh( dim, 1 )
+        .translate( { 0.0, 0.0, 1.0 - mesh_interpenetration } )
+        .updateAttrib( 1, 2 )
```
srwopschall: Again, can you add comment about the different integers here. I see that we have boundary attributes 6 and 7, but not sure where 1 and 8 come from, for instance.

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+  void SetUp3DCommonPlaneProblem( tribol::BinningMethod binning_method )
+  {
+    // binning proximity value
+    auto binning_proximity = std::get<1>( GetParam() );
+    // kinematic constant penalty stiffness
+    constexpr tribol::RealT penalty = 1.0;
+
+    auto mesh = SetUp3DMesh();
+
+    SetUpCommonPlaneProblem( std::get<0>( mesh ), std::get<1>( mesh ), std::get<2>( mesh ), penalty, binning_proximity,
```
srwopschall: Update function name once renamed for clarity.

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+                             binning_method );
+  }
+
+  void SetUpMortarProblem( tribol::BinningMethod binning_method )
```
srwopschall, line 171: Consider renaming consistent with how you rename SetupCommonPlaneProblem(). I realize there are no 2D and 3D variants for mortar. This routine kind of does it all. It Sets up the problem AND updates. I think I have something like this in TestMesh called SetupAndUpdate...

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+      EXPECT_GT( max_force_, 1.0e-6 );
+    } else {
+      EXPECT_LT( max_force_, 1.0e-15 );
+    }
+  }
+
+  MPI_Barrier( MPI_COMM_WORLD );
+}
+
+INSTANTIATE_TEST_SUITE_P(
```
srwopschall, line 257: You have it setup to check for max_force as a proxy for whether all the correct interface pairs have in fact been included. Perhaps add a comment that this is the point.

src/tests/tribol_proximity_check.cpp
```diff
@@ -0,0 +1,274 @@
...
+    tribol, ProximityTest,
+    testing::Values( std::make_tuple( 1, 0.0, 2.0, true ), std::make_tuple( 1, 0.0, 2.01, false ),
```
srwopschall: Add comment here as to whether you are testing for full inclusion of interface pairs, or whether they fall outside the binning proximity. It will help a user/developer if the set of inputs for each test has a comment as to what that test is looking for.

src/tribol/geom/ContactPlane.cpp
```diff
@@ -705,14 +680,10 @@ TRIBOL_HOST_DEVICE void ContactPlane::planePointAndCentroidGap( const MeshData::
...
   }
-  bool inPlane = false;
-  LinePlaneIntersection( xA, yA, zA, xB, yB, zB, m1.getElementCentroids()[0][fId1], m1.getElementCentroids()[1][fId1],
-                         c1_z, m1.getElementNormals()[0][fId1], m1.getElementNormals()[1][fId1], n1_z, xc1, yc1, zc1,
-                         inPlane );
-
-  LinePlaneIntersection( xA, yA, zA, xB, yB, zB, m2.getElementCentroids()[0][fId2], m2.getElementCentroids()[1][fId2],
-                         c2_z, m2.getElementNormals()[0][fId2], m2.getElementNormals()[1][fId2], n2_z, xc2, yc2, zc2,
-                         inPlane );
+  ProjectPointToPlane( xcg, ycg, zcg, m1.getElementNormals()[0][fId1], m1.getElementNormals()[1][fId1], n1_z,
```
srwopschall, line 683: Good. This is the better approach as discussed.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -69,6 +69,8 @@ void registerMfemCouplingScheme( IndexT cs_id, int mesh_id_1, int mesh_id_2, con
     }
   }
   coupling_scheme.setMfemMeshData( std::move( mfem_data ) );
+  // scale binning proximity if a higher order mesh is used
```
srwopschall: Is this comment correct? This looks like it sets it, but doesn't scale it.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -82,7 +84,19 @@ void setMfemLORFactor( IndexT cs_id, int lor_factor )
   SLIC_ERROR_ROOT_IF( !coupling_scheme->hasMfemData(),
                       "Coupling scheme does not contain MFEM data. "
                       "Create the coupling scheme using registerMfemCouplingScheme() to set the LOR factor." );
+
+  // reset binning proximity based on original LOR factor
```
srwopschall: Explain the rationale here for clarity. It gets confusing when considering HO to LO and scalings with LOR factor. Is the idea that the original binning proximity scale is some percent of the HO element thickness? Then, by dividing it by LOR factor it approximately becomes some percentage of the low order element thickness? Is that what we want? A user may want 2*HO_element_thickness to allow for one HO face to interpenetrate twice the thickness of the opposing HO element, for instance. By dividing by the lor_factor, they won't see contact with that amount of interpen? Since it is a scaling one approach isn't necessarily wrong, but I'm trying to think about what the user may expect when setting this. I think it is best if they don't have to think too much about the LOR mesh.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -82,7 +84,19 @@ void setMfemLORFactor( IndexT cs_id, int lor_factor )
...
+  auto orig_binning_proximity_scale = coupling_scheme->getParameters().binning_proximity_scale;
+  auto orig_lor_factor = coupling_scheme->getMfemMeshData()->GetLORFactor();
```
srwopschall: Perhaps I'm not seeing this correctly, but it looks like this is in the routine setMfemLORFactor(), which takes the lor_factor as input. Here, instead, you are getting it from the getter routine. Then, on line 96 you set the lor factor based on the input arg. Can you clarify?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -317,7 +331,7 @@ void updateMfemParallelDecomposition()
       // creates a new redecomp mesh based on updated coordinates and updates
       // transfer operators and displacement, velocity, and response grid
       // functions based on new redecomp mesh
-      mfem_data->UpdateMfemMeshData();
+      mfem_data->UpdateMfemMeshData( coupling_scheme.getParameters().binning_proximity_scale );
```
srwopschall: Why does the binning proximity scale need to be sent here. Shouldn't the binning proximity live on the coupling scheme and the coupling scheme drives mesh interactions? What is UpdateMfemMeshData() doing that needs the binning proximity scale?

src/tribol/interface/tribol.cpp
```diff
@@ -331,6 +331,17 @@ void setBinningProximityScale( IndexT cs_id, RealT binning_proximity_scale )
         "Setting binning proximity to less than 2.0 can lead to missed contact pairs.  Resetting to 2.0." );
   }
 
+#ifdef BUILD_REDECOMP
+
+  if ( cs->hasMfemData() ) {
+    auto lor_factor = cs->getMfemMeshData()->GetLORFactor();
+    if ( lor_factor > 0 ) {
+      binning_proximity_scale = binning_proximity_scale * static_cast<RealT>( lor_factor );
```
srwopschall: Can you clarify why we are not multiplying the binning proximity scale by the lor factor, but in another routine we divide it by the lor factor. This needs some comments to make this clear.

src/tribol/mesh/MfemData.cpp
```diff
@@ -295,7 +295,7 @@ void MfemMeshData::SetParentCoords( const mfem::ParGridFunction& current_coords
   coords_.SetParentGridFn( current_coords );
 }
 
-void MfemMeshData::UpdateMfemMeshData()
+void MfemMeshData::UpdateMfemMeshData( RealT binning_proximity_scale )
```
srwopschall, line 298: Not a big deal, but it looks a little funny to have this routine take the binning proximity scale. Seems like only the coupling scheme needs to know about the binning proximity scale. Is it because the binning scale needs to be modified by the lor factor, and that is internal to redecomp, not at the coupling scheme level, and therefore, the mfem mesh here needs to know about it?

src/tribol/mesh/MfemData.cpp
```diff
@@ -504,8 +504,16 @@ void MfemMeshData::SetMaterialModulus( mfem::Coefficient& modulus_field )
 MfemMeshData::UpdateData::UpdateData( mfem::ParSubMesh& submesh, mfem::ParMesh* lor_mesh,
                                       const mfem::ParFiniteElementSpace& parent_fes,
                                       mfem::ParGridFunction& submesh_gridfn, SubmeshLORTransfer* submesh_lor_xfer,
-                                      const std::set<int>& attributes_1, const std::set<int>& attributes_2 )
-    : redecomp_mesh_{ lor_mesh ? redecomp::RedecompMesh( *lor_mesh ) : redecomp::RedecompMesh( submesh ) },
+                                      const std::set<int>& attributes_1, const std::set<int>& attributes_2,
+                                      RealT binning_proximity_scale )
+    : redecomp_mesh_{ lor_mesh ? redecomp::RedecompMesh(
```
srwopschall, line 509: Make sure you have a comment somewhere about what the binning scale * max element size is for. I'm assuming this is for creating a max bounding box dimension for RCB or something like that?

src/tribol/mesh/MfemData.hpp
```diff
@@ -988,6 +990,13 @@ class MfemMeshData {
     return submesh_lor_xfer_ ? submesh_lor_xfer_->GetLORGridFn().ParFESpace() : nullptr;
   }
 
+  /**
+   * @brief Get the number of element subdivisions per dimension on the LOR mesh
```
srwopschall: Minor, but the number of subdivisions is equal in all directions across the HO face, right?

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -610,27 +612,27 @@ class BvhSearch : public SearchBase {
...
+                    }
+                    box.addPoint( pos );
                   }
                   // Expand the bounding box in the face normal direction
                   RealT vnorm[3];
                   mesh.getFaceNormal( i, vnorm );
                   VectorT faceNormal( vnorm );
                   RealT faceRadius = mesh.getFaceRadius()[i];
-                  expandBBoxNormal( box, faceNormal, faceRadius );
+                  expandBBoxNormal( box, faceNormal, binning_proximity * faceRadius );
```
srwopschall, line 640: I see that faceRadius was always used here. Perhaps this is still what we want, but we also should have the meshes element thicknesses associated with the faces. Would that be more appropriate? At the time we only had the face radius as a proximity length scale. The potential issue here is that for thin but long elements, the face radius will be more inclusive for adjacent and nearly adjacent elements. For really small faces but deep elements, the element thickness will be more inclusive. Thoughts?

srwopschall (commented):
@ebchin - Thanks for pushing this change through the rest of the code. The MeshBuilder() looks clean and useful. I do have some comments that need to be addressed or discussed prior to approval. Some are minor, some may require short discussion.

src/shared/mesh/MeshBuilder.cpp
ebchin: Merge seems OK to me.  Maybe Join or Union?  For what you describe, I'd use something like "Stitch", since it's more than just putting the two meshes together in one data structure, it's modifying the connectivity to tie them together.

src/shared/CMakeLists.txt
ebchin: That's correct, redecomp is below tribol (i.e. redecomp is a tribol dependency), but we can build tribol without redecomp.  The only way for tribol to see MeshBuilder always is to do a separate library.  Other things should belong in shared so the hierarchy is consistent (stuff used in tribol's config.hpp is used in redecomp), but they haven't been enforced since shared didn't exist.  I plan to move that stuff over in a future PR.

src/shared/mesh/MeshBuilder.hpp
ebchin: I prefer this syntax, but it resulted in a few more move constructors being called in clang.  I left it here as an option for now.

src/shared/mesh/MeshBuilder.hpp
ebchin: I'd say start with a unit square and then use transformations to get the shape you want.  It keeps each component as simple as possible.

src/shared/mesh/MeshBuilder.hpp
ebchin: The idea is if we want a dimension parameter in a test or a function.  This will let you pass that dimension on to the mesh constructor.  I wasn't able to use it here because we had to change attributes in this test, but it might be useful when this isn't needed.  I'd like to add something like picking attributes based on a bounding box, etc. so it's independent of boundary attributes.

src/shared/mesh/MeshBuilder.hpp
ebchin: Potentially.  I think this could be another method we could add later on.

src/shared/mesh/MeshBuilder.hpp
ebchin: Yes, it translates the whole mesh.  I only needed translation for this test.  If other tests/examples need scaling, etc. I can add them in the future.

src/shared/mesh/MeshBuilder.cpp
ebchin: I'll add a comment

src/tests/tribol_proximity_check.cpp
ebchin: That's what the print boundary attributes method is for, so you know where the boundary attributes are.  I don't see any documentation in mfem's mesh creation methods, so I don't think we should add any notes etc. here because it isn't defined in mfem's documentation what the attribute sets are, so they could change.  This is why I think picking attributes based on a bounding box might be more useful and general.

src/tests/tribol_proximity_check.cpp
ebchin: What are your thoughts on getting rid of this all together?  I did it this way to get rid of code duplication, but putting everything in the test itself might be clearer as to what's going on (less scrolling).

src/tribol/interface/mfem_tribol.cpp
ebchin: This stuff could be made cleaner, but the problem is the binning proximity is set in Tribol but the LOR factor is set in the MFEM interface (which sits above Tribol).  Maybe LOR factor could be moved down to the Tribol level (just to set binning proximity cleaner)?

src/tribol/interface/mfem_tribol.cpp
ebchin: The redecomp mesh needs the binning proximity.  It's created in this method.

src/tribol/mesh/MfemData.cpp
ebchin: This is for creating the redecomp mesh, which doesn't know anything about a coupling scheme.

src/tribol/mesh/MfemData.hpp
ebchin: Correct

src/tribol/search/InterfacePairFinder.cpp
ebchin: I think face radius is fine; there isn't always a notion of thickness at the Tribol mesh level (unless the field is registered with Tribol).

src/shared/mesh/MeshBuilder.cpp
ebchin: I think this might be OK since it says boundary on the line above?

src/shared/mesh/MeshBuilder.cpp
ebchin: Comment added

src/tests/tribol_proximity_check.cpp
ebchin: It's being used as a verb here, so "set up" is correct.  See e.g. https://grammarist.com/spelling/set-up-vs-setup/

src/tests/tribol_proximity_check.cpp
ebchin: Yeah, it's similar to the SetupAndUpdate routine, though for the Tribol MFEM API.  It might be good to break this out into a separate file too, though in a follow on PR.

ebchin: On second thought, I think I'll just get rename things for now using your suggestions.  If we decide on a better way to "drive" Tribol in this example, we can change it later.

src/shared/mesh/MeshBuilder.cpp
ebchin: I added a clarifying note to the constructors

src/tests/tribol_proximity_check.cpp
ebchin: Comment added

src/tests/tribol_proximity_check.cpp
ebchin: Added a comment describing what the test parameters are.

src/tribol/interface/mfem_tribol.cpp
ebchin: I think moving LOR factor to a Tribol parameter makes sense, just so we can clean up this code.

src/shared/mesh/MeshBuilder.cpp
ebchin: Change to Unify

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -832,6 +839,8 @@ class CouplingScheme {
 
   PairReportingData m_pairReportingData;  ///< struct handling on-rank pair reporting data from computational geometry
 
+  RealT m_effective_binning_proximity_scale;  ///< Binning proximity scaled by the LOR factor
```
srwopschall: Minor: consider adding note as to why we do this. Scaling by the LOR factor makes the proximity detection independent of the LO refinement so you get consistent RCB, binning, geom filter for a fixed HO mesh regardless of LOR factor. I can envision forgetting this unless it is written down

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -210,7 +210,7 @@ class CartesianProduct : public SearchBase {
     ContactMode cmode = m_coupling_scheme->getContactMode();
 
     bool auto_contact_check = m_coupling_scheme->getParameters().auto_contact_check;
-    auto binning_proximity_scale = m_coupling_scheme->getParameters().binning_proximity_scale;
+    auto binning_proximity_scale = m_coupling_scheme->getEffectiveBinningProximityScale();
```
srwopschall: I think this is ok because it is a local variable inside the grid binning, but potentially confusing. I first looked at this thinking that we should be setting `auto binning_proximity_scale` to the binning proximity scale stored on the coupling scheme, but recalled that we always want to be using the effective binning scale. Just as long as it is clear that we actually want to be using the effective.

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -325,7 +325,7 @@ class GridSearch : public SearchBase {
     const RealT bboxTolerance = 1e-6;
 
     m_coupling_scheme->getInterfacePairs().clear();
-    auto binning_proximity_scale = m_coupling_scheme->getParameters().binning_proximity_scale;
+    auto binning_proximity_scale = m_coupling_scheme->getEffectiveBinningProximityScale();
```
srwopschall: Same comment about potential conflict of naming and clarity issue.

srwopschall (approved):
@ebchin - Great. A few minor comments for clarity.

src/tribol/search/InterfacePairFinder.cpp
ebchin: Right, I can see why this might be confusing.  I added a note to the `getEffectiveBinningProximityScale()` that should clear up its usage.

src/tribol/mesh/CouplingScheme.hpp
ebchin: Added

PR #129, opened 2025-02-03
Get rid of braced initializer for `mfem::Vector` and `mfem::Array<T>`

ebchin:
A `std::initializer_list` constructor was recently added to `mfem::Vector`, so code like `mfem::Array<int> blah {100};` now creates a `mfem::Array<int>` of size 1 with `blah[0] = 100`, instead of an uninitialized `mfem::Array<int>` of size 100.

This PR updates the places in the code where this is an issue.  Note, the verification of changes is through running tests and there may be other places where the code still needs to be updated.

### DIFF
--- a/src/tribol/mesh/MfemData.cpp
+++ b/src/tribol/mesh/MfemData.cpp
@@ -661,7 +661,7 @@ const MfemSubmeshData::UpdateData& MfemSubmeshData::GetUpdateData() const
 
 MfemJacobianData::MfemJacobianData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data,
                                     ContactMethod contact_method )
-    : parent_data_{ parent_data }, submesh_data_{ submesh_data }, block_offsets_{ 3 }
+    : parent_data_{ parent_data }, submesh_data_{ submesh_data }, block_offsets_( 3 )
 {
   SLIC_ERROR_ROOT_IF( parent_data.GetParentCoords().ParFESpace()->FEColl()->GetOrder() > 1,
                       "Higher order meshes not yet supported for Jacobian matrices." );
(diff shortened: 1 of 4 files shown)


src/tribol/mesh/MfemData.cpp
```diff
@@ -609,7 +609,7 @@ void MfemMeshData::UpdateData::SetElementData()
     num_verts_per_elem_ = mfem::Geometry::NumVerts[element_type];
   } else {
     // just put something here so Tribol will not give a warning for zero element meshes
-    elem_type_ = LINEAR_EDGE;
+    elem_type_ = LINEAR_QUAD;
```
srwopschall, line 613: Perhaps enhance the comment to currently put the surface topology (we don't support volume interfaces yet) with the max number of nodes?

src/tribol/mesh/MfemData.cpp
```diff
@@ -661,7 +661,7 @@ const MfemSubmeshData::UpdateData& MfemSubmeshData::GetUpdateData() const
 
 MfemJacobianData::MfemJacobianData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data,
                                     ContactMethod contact_method )
-    : parent_data_{ parent_data }, submesh_data_{ submesh_data }, block_offsets_{ 3 }
+    : parent_data_{ parent_data }, submesh_data_{ submesh_data }, block_offsets_( 3 )
```
srwopschall, line 665: Comment: It is what it is, but as we talked about this is kind of annoying since you see a mix of braces and then the parenthesis. If you didn't know, you'd have to go to the header to differentiate between the variables and see that mfem has the different constructor convention. I don't think this warrants a comment or anything, but it could be very easy to miss if you didn't know.

srwopschall (approved):
@ebchin - This was shorter than I thought, so easy to review. If you haven't yet, perhaps as a sanity check, just search for any mfem::Vector in our source and then search for those variable names to see if you got everything?

src/tribol/mesh/MfemData.cpp
ebchin: Agreed, but we might be able to clean some of this up in a future PR.

PR #132, opened 2025-02-26
Task/fromm3/penalty check

jefromm:


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -441,6 +433,14 @@ bool CouplingScheme::isValidCouplingScheme()
       break;
   }
 
+  if ( !this->isValidEnforcement() ) {
```
srwopschall, line 436: Looks like it probably doesn't matter, but did this if-block need to be relocated after the switch statement on `checkExecutionModeData()`?

srwopschall (commented):
@jefromm - Thanks for tackling this; this looks great. I want @ebchin to take a quick look and approve since this is likely going to be the model used elsewhere in the code (as needed/if needed).

src/tribol/mesh/CouplingScheme.cpp
ebchin: It looks like `m_exec_mode` is set in `checkExecutionModeData()`.  We probably shouldn't be setting anything in `isValidCouplingScheme()` (i.e. this method should be `const`), but then we would need to move the execution mode logic out of this method.  Probably best to leave it as is for now.

ebchin (approved):
Thanks @jefromm.  Looks great!  We should add `KINEMATIC_ELEMENT` to the `tribol_common_plane_gpu.cpp` test, but we can get this in for now.

PR #137, opened 2025-03-18
fixes for no raja and no mpi builds

ebchin:


jefromm (approved):

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -131,6 +131,8 @@ MeshBuilder::operator mfem::Mesh&() { return mesh_; }
 
 MeshBuilder::operator const mfem::Mesh&() const { return mesh_; }
 
+#ifdef TRIBOL_USE_MPI
```
srwopschall, line 134: Looks like this will enable a serial build, but if a serial build does not use MPI does this break any tests?

srwopschall (approved):
@ebchin, @jefromm - Looks good. One comment for clarification.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1287,8 +1287,10 @@ void CouplingScheme::computeCommonPlaneTimeStep( RealT& dt )
   ArrayT<RealT> dt_temp_data( { dt, dt }, getAllocatorId() );
   ArrayViewT<RealT> dt_temp = dt_temp_data;
   // [0]: exceed_max_gap1, [1]: exceed_max_gap2, [2]: neg_dt_gap_msg, [3]: neg_dt_vel_proj_msg
-  ArrayT<bool> msg_data( { false, false, false, false }, getAllocatorId() );
-  ArrayViewT<bool> msg = msg_data;
+  ArrayT<IndexT> msg_data( { static_cast<IndexT>( false ), static_cast<IndexT>( false ), static_cast<IndexT>( false ),
```
srwopschall, line 1290: @ebchin This looks good to me. What do you think?

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1464,8 +1466,13 @@ void CouplingScheme::computeCommonPlaneTimeStep( RealT& dt )
                   dt1_check1 = ( dt1_vel_check ) ? exceed_max_gap1 : false;
                   dt2_check1 = ( dt2_vel_check ) ? exceed_max_gap2 : false;
 
+#ifdef TRIBOL_USE_RAJA
+                  RAJA::atomicMax<RAJA::auto_atomic>( &msg[0], static_cast<IndexT>( exceed_max_gap1 ) );
```
srwopschall, line 1470: I think this should be ok right? The entries in `msg` are initialized to the static cast of false, so 0. We are interested in anytime they might be triggered to true, which should be static cast to 1 here, so atomic max makes sense.

srwopschall (commented):
I think should be good. Two comments for sanity check.

src/shared/mesh/MeshBuilder.cpp
ebchin: Serial builds cannot use MPI, so `ParMeshBuilder` should not be exposed to them

PR #139, opened 2025-03-27
64-bit integers for `axom::IndexType`

ebchin:
- Changes required to get 64-bit integer support for `axom::IndexType` (and, by extension, `tribol::IndexT`).
- Fix some bugs in `redecomp::MPIUtility` when transferring `axom::Array`s
- Expose the number of ranks in the contact redecomposition to the MFEM interface
- Enable int64 support in our axom spack recipe

Note we are not testing 64 bit integer support in CI.  See issue #175.


scripts/spack/packages/axom/package.py
```diff
@@ -14,3 +14,25 @@ class Axom(BuiltinAxom):
     # Note: We add a number to the end of the real version number to indicate that we have
     #  moved forward past the release. Increment the last number when updating the commit sha.
     version("0.10.1.1", commit="44562f92a400204e33915f48b848eb68e80a1bf1", submodules=False)
+
+    variant("int64", default=True, description="Use 64bit integers for index type")
```
srwopschall, line 18: Aside from simply needing this in a given host-code, what are the consequences to defaulting to True here?

src/redecomp/partition/RCB.cpp
```diff
@@ -29,23 +29,29 @@ std::vector<EntityIndexByRank> RCB<NDIMS>::generatePartitioning(
   // Build a partitioning using recursive coordinate bisection
   auto problem_tree = BuildProblemTree( n_parts, coords_by_mesh, ghost_len );
 
+  // Ensures ranks with elements are evenly distributed throughout the partition
```
srwopschall: Consider rewording? This reads like the ranks are evenly distributed, which I don't follow. I think we want the elements to be evenly distributed across the selected number of non-zero-element ranks.

src/redecomp/partition/RCB.cpp
```diff
@@ -29,23 +29,29 @@ std::vector<EntityIndexByRank> RCB<NDIMS>::generatePartitioning(
...
-      ent_ghost[i].reserve( coords.size() );
+      auto rank = i * rank_multiplier;
+      ent_idx[rank].reserve( coords.size() );
+      ent_ghost[rank].reserve( coords.size() );
     }
     for ( int i{ 0 }; i < coords.size(); ++i ) {
       auto dest = DetermineDomain( problem_tree, coords[i] );
-      ent_idx[dest].push_back( i );
-      ent_ghost[dest].push_back( false );
+      auto rank = dest * rank_multiplier;
```
srwopschall: is this guaranteed to be an integer? I'm wondering because rank_multiplier isn't necessarily a whole number? "rank" is then used to index into the array.

src/redecomp/partition/RCB.cpp
```diff
@@ -29,23 +29,29 @@ std::vector<EntityIndexByRank> RCB<NDIMS>::generatePartitioning(
...
+      ent_idx[rank].push_back( i );
+      ent_ghost[rank].push_back( false );
       const auto& neighbors = problem_tree( dest ).neighbor_bboxes_;
       for ( int j{ 0 }; j < neighbors.size(); ++j ) {
         if ( problem_tree( neighbors[j] ).ghost_bbox_.contains( coords[i] ) ) {
-          ent_idx[neighbors[j]].push_back( i );
-          ent_ghost[neighbors[j]].push_back( true );
+          auto neighbor_rank = neighbors[j] * rank_multiplier;
```
srwopschall: same comment about using neighbor_rank to index into the array, but is neighbor_rank a whole number?

src/redecomp/partition/RCB.cpp
srwopschall: Let's discuss that rank_multiplier. I'm not seeing exactly what it is trying to do.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -48,6 +50,7 @@ mfem::SparseMatrix MatrixTransfer::TransferToParallelSparse( const axom::Array<i
                                                              const axom::Array<int>& trial_elem_idx,
                                                              const axom::Array<mfem::DenseMatrix>& src_elem_mat ) const
 {
+  // TODO: we need a SparseMatrix-like data structure that allows HYPRE_BigInt on the columns
```
srwopschall: if this remark carries enough weight consider making it an issue with a little explanation.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -147,10 +150,21 @@ std::unique_ptr<mfem::HypreParMatrix> MatrixTransfer::ConvertToHypreParMatrix( m
   SLIC_ERROR_IF( sparse.Width() != parent_trial_fes_.GlobalVSize(),
                  "Width of sparse must match number of trial ParFiniteElementSpace global dofs." );
 
+  // TODO: mfem::SparseMatrix uses int for global column values; this needs to be HYPRE_BigInt to be compatible with
```
srwopschall: Same comment re elevating this to an issue?

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -400,11 +414,17 @@ MPIArray<int> MatrixTransfer::buildRecvTrialElemDofs( const RedecompMesh& trial_
               elem_id = trial_p2r_elems.first[src_trial_r][elem_offset];
               auto n_elem_dofs = parent_trial_fes_.GetFE( elem_id )->GetDof() * parent_trial_fes_.GetVDim();
               trial_dofs.resize( trial_dofs.size() + n_elem_dofs );
-              auto dof_array = mfem::Array<int>( &trial_dofs[trial_dofs.size() - n_elem_dofs], n_elem_dofs );
-              parent_trial_fes_.GetElementVDofs( elem_id, dof_array );
-              for ( auto& dof : dof_array ) {
-                dof += first_dof;
+              auto dof_array = mfem::Array<HYPRE_BigInt>( &trial_dofs[trial_dofs.size() - n_elem_dofs], n_elem_dofs );
+              auto int_dof_array = mfem::Array<int>( n_elem_dofs );
```
srwopschall, line 418: This is in the weeds, but maybe a comment that mfem won't return a BigInt array, and you have to do this workaround?

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -400,11 +414,17 @@ MPIArray<int> MatrixTransfer::buildRecvTrialElemDofs( const RedecompMesh& trial_
...
+              parent_trial_fes_.GetElementVDofs( elem_id, int_dof_array );
+              for ( int i{ 0 }; i < n_elem_dofs; ++i ) {
+                dof_array[i] = first_dof + static_cast<HYPRE_BigInt>( int_dof_array[i] );
               }
+              // auto dof_array_write = dof_array.HostWrite();
```
srwopschall, line 424: Clarify if we still need this commented out code. If so, add comment as to why it is still there.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -419,7 +439,7 @@ MPIArray<int> MatrixTransfer::buildRecvTrialElemDofs( const RedecompMesh& trial_
     send_trial_elem_dofs[test_r].reserve( send_trial_dof_sizes[test_r] );
     auto n_trial_elems = send_trial_elem_rank[test_r].size();
     for ( int e{ 0 }; e < n_trial_elems; ++e ) {
-      auto trial_elem_dofs = axom::ArrayView<int>(
+      auto trial_elem_dofs = axom::ArrayView<HYPRE_BigInt>(
```
srwopschall, line 443: Commenting randomly here, but it looks like a simple find and replace swapping an `int` for a `HYPRE_BigInt` wouldn't really work? Seems easy to miss swapping one for the other? Have you combed through this carefully?

src/redecomp/transfer/MatrixTransfer.hpp
```diff
@@ -152,10 +152,11 @@ class MatrixTransfer {
    * @param trial_redecomp Redecomp mesh of the trial space
    * @param test_elem_idx List of element IDs on the redecomp test space
    * @param trial_elem_idx List of element IDs on the redecomp trial space
-   * @return MPIArray<int>
+   * @return MPIArray<HYPRE_BigInt>
    */
-  MPIArray<int> buildRecvTrialElemDofs( const RedecompMesh& trial_redecomp, const axom::Array<int>& test_elem_idx,
-                                        const axom::Array<int>& trial_elem_idx ) const;
+  MPIArray<HYPRE_BigInt> buildRecvTrialElemDofs( const RedecompMesh& trial_redecomp,
+                                                 const axom::Array<int>& test_elem_idx,
```
srwopschall, line 158: Can you clarify why these are both `<int>` rather than `HYPRE_BigInt`?

src/redecomp/utils/MPIUtility.hpp
```diff
@@ -310,7 +310,7 @@ void MPIUtility::Send( const T& container, int dest, int tag ) const
 template <typename T, axom::MemorySpace Sp>
 void MPIUtility::Send( const axom::Array<T, 2, Sp>& container, int dest, int tag ) const
 {
-  MPI_Send( container.shape().m_data, 2, GetMPIType<int>(), dest, tag, comm_ );
+  MPI_Send( container.shape().m_data, 2, GetMPIType<axom::IndexType>(), dest, tag, comm_ );
```
srwopschall, line 313: Can you clarify the use of axom::IndexType here? We once had our own IndexType. Does axom's handle the 64 bit integer issue under the hood because we are building with that variant? This is compatible with `long` and `long long` used in the MPI Utility functions?

src/redecomp/utils/MPIUtility.hpp
```diff
@@ -349,8 +349,8 @@ template <typename T, axom::MemorySpace Sp>
 axom::Array<T, 2, Sp> MPIUtility::Recv( type<axom::Array<T, 2, Sp>>, int source, int tag ) const
 {
   auto container = axom::Array<T, 2, Sp>();
-  axom::StackArray<int, 2> dim_size;
-  MPI_Recv( dim_size.m_data, 2, GetMPIDatatype( container.data() ), source, tag, comm_, &status_ );
```
srwopschall, line 353: Is GetMPIDatatype() no longer needed or used anywhere?

src/tests/tribol_coupling_scheme.cpp
```diff
@@ -48,7 +48,7 @@ class CouplingSchemeTest : public ::testing::Test {
  public:
   int m_numCells;
   int m_lengthNodalData;
-  int* m_connectivity{ nullptr };
+  tribol::IndexT* m_connectivity{ nullptr };
```
srwopschall, line 51: Can you clarify use of tribol::IndexT here versus axom::IndexType used in other files?

src/tests/tribol_coupling_scheme.cpp
```diff
@@ -594,8 +594,8 @@ TEST_F( CouplingSchemeTest, invalid_mesh_in_coupling_scheme )
   int numFaces2 = 1;
   constexpr int numTotalNodes1 = 2;
```
srwopschall, line 595: Can you clarify when using int is ok, and when we should use tribol::IndexT? I can see myself developing and easily breaking a host-code built or ability to run on the gpu.

src/tests/tribol_enzyme_poly_intersect.cpp
```diff
@@ -49,9 +49,9 @@ class EnzymePolyIntersectTest : public testing::Test {
     // EdgeEdge = intersection of edges
     OverlapVertexType type[max_overlap_vert];
     // this stores which edge or vertex on element 1 the overlap vertex is associated with
-    IndexT edge1[max_overlap_vert];
+    int edge1[max_overlap_vert];
```
srwopschall, line 52: Why is this array int and not tribol::IndexT?

src/tests/tribol_mortar_data_geom.cpp
```diff
@@ -106,11 +106,16 @@ TEST_F( MortarGeomTest, mortar_good_patch )
...
 
-  this->v_ixm.SetSize( this->lengthMortarConn );
-  this->v_ixs.SetSize( this->lengthNonmortarConn );
+  this->v_ixm.resize( this->lengthMortarConn );
+  for ( int i{ 0 }; i < this->lengthMortarConn; ++i ) {
+    i_ixm >> v_ixm[i];
+  }
+
+  this->v_ixs.resize( this->lengthNonmortarConn );
+  for ( int i{ 0 }; i < this->lengthNonmortarConn; ++i ) {
```
srwopschall, line 116: I'm assuming these loops are doing what `Load()` did? Hence removing those calls? I don't see how i_ixm was set before. Perhaps add a comment clarifying what is being done here.

src/tests/tribol_mortar_data_geom.cpp
```diff
@@ -202,11 +207,16 @@ TEST_F( MortarGeomTest, mortar_bad_patch )
   std::ifstream i_ys( ys_file );
   std::ifstream i_zs( zs_file );
 
-  this->v_ixm.SetSize( this->lengthMortarConn );
-  this->v_ixs.SetSize( this->lengthNonmortarConn );
+  this->v_ixm.resize( this->lengthMortarConn );
+  for ( int i{ 0 }; i < this->lengthMortarConn; ++i ) {
```
srwopschall, line 213: Consider adding a quick comment here and elsewhere in this test where these manual loops are doing these shifts.

src/tests/tribol_mortar_data_weights.cpp
```diff
@@ -150,11 +137,16 @@ TEST_F( MortarSparseWtsTest, mortar_sphere )
   std::ifstream i_ys( ys_file );
   std::ifstream i_zs( zs_file );
 
-  this->v_ixm.SetSize( this->lengthMortarConn );
-  this->v_ixs.SetSize( this->lengthNonmortarConn );
+  this->v_ixm.resize( this->lengthMortarConn );
+  for ( int i{ 0 }; i < this->lengthMortarConn; ++i ) {
```
srwopschall, line 142: Comment here and throughout at these manual loops?

src/tests/tribol_nodal_nrmls.cpp
```diff
@@ -37,7 +37,7 @@ class NodalNormalTest : public ::testing::Test {
   int dim;
 
   void computeNodalNormals( int cell_type, RealT const* const x, RealT const* const y, RealT const* const z,
-                            int const* const conn, int const numCells, int const numNodes, int const dim )
+                            const tribol::IndexT* conn, int const numCells, int const numNodes, int const dim )
```
srwopschall, line 40: Confirm removing second `const`

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -369,7 +369,7 @@ mfem::ParGridFunction& getMfemPressure( IndexT cs_id )
   return cs->getMfemSubmeshData()->GetSubmeshPressure();
 }
 
-void updateMfemParallelDecomposition()
+void updateMfemParallelDecomposition( int n_ranks )
```
srwopschall, line 372: It is common for this routine to be called cycle to cycle in an explicit code. Is there a disadvantage to potentially changing the number of ranks argument between cycle calls? Do we want to guard against this? Do we only want to decompose over a fixed set of ranks for a given coupling scheme as long as that coupling scheme persists?

src/tribol/mesh/MfemData.cpp
```diff
@@ -591,8 +594,8 @@ void MfemMeshData::UpdateData::UpdateConnectivity( const std::set<int>& attribut
                                                    const std::set<int>& attributes_2 )
 {
   // create this on host since MFEM connectivity data is stored there
-  Array2D<int, MemorySpace::Host> conn_1_host;
-  Array2D<int, MemorySpace::Host> conn_2_host;
+  Array2D<IndexT, MemorySpace::Host> conn_1_host;
+  Array2D<IndexT, MemorySpace::Host> conn_2_host;
   Array1D<int, MemorySpace::Host> elem_map_1_host;
```
srwopschall, line 599: Why not make this `int` an IndexT? Elsewhere for a face id, for instance, you switched `int` -> `IndexT`. I'm just trying to understand the rationale.

src/tribol/mesh/MfemData.cpp
```diff
@@ -632,8 +635,8 @@ void MfemMeshData::UpdateData::UpdateConnectivity( const std::set<int>& attribut
     elem_map_2_ = std::move( elem_map_2_host );
   } else {
     // copy to new memory space
-    conn_1_ = Array2D<int>( conn_1_host, allocator_id_ );
-    conn_2_ = Array2D<int>( conn_2_host, allocator_id_ );
+    conn_1_ = Array2D<IndexT>( conn_1_host, allocator_id_ );
+    conn_2_ = Array2D<IndexT>( conn_2_host, allocator_id_ );
     elem_map_1_ = Array1D<int>( elem_map_1_host, allocator_id_ );
```
srwopschall, line 640: Similar comment about `int` -> `IndexT` for the element map.

src/tribol/mesh/MfemData.cpp
```diff
@@ -871,8 +877,12 @@ std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian( con
       GetUpdateData().submesh_redecomp_xfer_10_->TransferToParallelSparse( lm_elems, nonmortar_elems, *elem_J_2 );
   submesh_J.Finalize();
 
-  // transform J values from submesh to parent mesh
-  auto J = submesh_J.GetJ();
+  // transform J values from submesh to (global) parent mesh
+  mfem::Array<HYPRE_BigInt> J( submesh_J.NumNonZeroElems() );
```
srwopschall, line 881: For consistency consider adding a note here about the int to BigInt workaround.

src/tribol/mesh/MfemData.cpp
```diff
@@ -945,7 +955,7 @@ std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian( con
   // trial space is on the parent mesh, not the submesh
   auto J_full = std::make_unique<mfem::HypreParMatrix>( mpi.MPIComm(), submesh_fes.GetVSize(),
                                                         submesh_fes.GlobalVSize(), parent_trial_fes.GlobalVSize(),
-                                                        submesh_J.GetI(), submesh_J.GetJ(), submesh_J.GetData(),
+                                                        submesh_J.GetI(), J.GetData(), submesh_J.GetData(),
```
srwopschall, line 960: This looks like it loses a little naming consistency that could lead to confusion. Is `J` on the submesh? Is it not? You have two arrays now instead of just the one. Consider appending "submesh" to the name and add a comment to why you can no longer just call `submesh_J.GetJ()`.

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -205,7 +205,7 @@ class CartesianProduct : public SearchBase {
     IndexT mesh2NumElems = mesh2.numberOfElements();
 
     // Reserve memory for boolean array indicating which pairs are proximate
-    int maxNumPairs = mesh1NumElems * mesh2NumElems;
```
srwopschall, line 208: I made this comment elsewhere, but seems like it would be easy to miss changing an `int` to `IndexT`, and that even some judgement has to be made as to whether you even need to do it. Have you reviewed these cases relatively thoroughly?

src/tribol/utils/Math.cpp
```diff
@@ -90,15 +90,17 @@ int binary_search( const int* const array, const int n, const int val )
...
 {
   int temp = *xp;
   *xp = *yp;
   *yp = temp;
 }
 
 //------------------------------------------------------------------------------
-void bubble_sort( int* const array, const int n )
+template <typename T>
+void bubble_sort( T* array, int n )
```
srwopschall, line 103: Keep `int` for `n` in the argument list? Seems like n could be very large?

src/tribol/utils/Math.cpp
```diff
@@ -110,6 +112,10 @@ void bubble_sort( int* const array, const int n )
   }
 }
 
+template void bubble_sort( int* array, int n );
+template void bubble_sort( long* array, int n );
```
srwopschall, line 116: Do we have alternatives other than `long` and `long long` here that might be more consistent with other parts of the code?

src/tribol/utils/Math.cpp
```diff
@@ -170,6 +176,19 @@ void allocIntArray( int** arr, const int length, const int* const data )
...
+void allocArray( T** arr, int length, T init_val )
+{
+  SLIC_ERROR_IF( length == 0, "allocIntArray: please specify nonzero length "
+                                  << "for array allocation." );
+
+  *arr = new T[length];
+  initArray( *arr, length, init_val );
+}
+
+template void allocArray( IndexT** arr, int length, IndexT init_val );
```
srwopschall, line 190: Why not do IndexT length? Just curious.

srwopschall (commented):
Some minor comments. Open to discussing to speed this up.

src/redecomp/partition/RCB.cpp
ebchin: Yes

src/redecomp/partition/RCB.cpp
ebchin: Do a simple example in the comment

ebchin: Rename `rank_jump_interval`

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: Add name to comment

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: I think this is what we would need on GPU.  Add a comment explaining this.

src/tests/tribol_mortar_data_geom.cpp
ebchin: Comment: This replicates the Load() method since mfem::Array doesn't support 64-bit integers.

src/tests/tribol_mortar_data_geom.cpp
ebchin: Same comment

src/tests/tribol_mortar_data_geom.cpp
```diff
@@ -298,11 +308,16 @@ TEST_F( MortarGeomTest, mortar_ironing )
   std::ifstream i_ys( ys_file );
   std::ifstream i_zs( zs_file );
 
-  this->v_ixm.SetSize( this->lengthMortarConn );
-  this->v_ixs.SetSize( this->lengthNonmortarConn );
+  this->v_ixm.resize( this->lengthMortarConn );
```
ebchin, line 314: Same comment

src/tests/tribol_mortar_data_geom.cpp
```diff
@@ -424,11 +439,16 @@ TEST_F( MortarGeomTest, mortar_ironing_block_sub_mesh )
...
```
ebchin, line 446: Same comment

src/tribol/mesh/MfemData.cpp
ebchin: Rename `J` to `bigint_submesh_J`

src/redecomp/utils/MPIUtility.hpp
ebchin: It's used elsewhere. This was a bugfix.

src/tribol/mesh/MfemData.cpp
ebchin: Looking more closely `J` is fine here.  The columns need to be HYPRE_BigInt DOFs on the parent mesh, not the submesh.

src/tribol/mesh/MfemData.cpp
ebchin: Comment added

srwopschall (approved):

PR #140, opened 2025-04-02
setGapSeparationRatio function and cmake fix

thartland:
Adding a setGapSeparationRatio utility function and fixing a bug in a cmake file.


ebchin (approved):
Thanks for this @thartland, looks good to me!

srwopschall:
Closing PR. Will bring over necessary CMake changes and address the gap separation ratio in a separate PR.

PR #141, opened 2025-04-15
Add TRIBOL_DEBUG and add guards around SLIC calls on device

ebchin:
Added a `TRIBOL_DEBUG` compiler definition (lifted heavily from the one in axom) and added some `TRIBOL_ON_HOST` compiler guards for SLIC macros in host/device code.


CMakeLists.txt
```diff
@@ -128,9 +128,44 @@ if (TRIBOL_STYLE_CI_ONLY)
 endif()
 
 #------------------------------------------------------------------------------
-# Add TPLs and source directories
+# Add TPLs
 #------------------------------------------------------------------------------
 include(cmake/SetupThirdParty.cmake)
+
```
ebchin, line 134: @chapman39 This block will have to be added to the `SetupSeracThirdParty.cmake` file for codevelop builds.  Would it be better to put this in a file in the `cmake` folder (that's how axom does it)?

CMakeLists.txt
```diff
@@ -128,9 +128,44 @@ if (TRIBOL_STYLE_CI_ONLY)
...
+  else () # This case handles multi-config generators, e.g. MSVC
+      set(TRIBOL_DEBUG_DEFINE_STRING "$<$<CONFIG:Debug,RelWithDebInfo>:TRIBOL_DEBUG>")
+  endif()
+else()  # Handle bad value for TRIBOL_DEBUG_DEFINE config variable
+  message(FATAL_ERROR 
+    "Invalid value for TRIBOL_DEBUG_DEFINE. Must be 'DEFAULT', 'ON' or 'OFF'; was '${TRIBOL_DEBUG_DEFINE}'")
+endif()
+
+set(TRIBOL_DEBUG_DEFINE_STRING "${TRIBOL_DEBUG_DEFINE_STRING}" CACHE STRING "" FORCE)
+# Define AXOM_DEBUG for SLIC macros
```
srwopschall, line 162: Is this bit here just to make sure that if tribol debug is on, then we set axom debug accordingly?

src/tribol/integ/FE.cpp
```diff
@@ -101,6 +101,7 @@ TRIBOL_HOST_DEVICE void SegmentBasis( const RealT* const x, const RealT pX, cons
 
   phi = 1.0 / lambda * ( lambda - magW );
 
+#ifdef TRIBOL_USE_HOST
```
srwopschall, line 104: do we also want to make these slic_debug statements contingent of if defined tribol_debug? they look like debug statements.

src/tribol/integ/Integration.cpp
```diff
@@ -79,19 +79,23 @@ TRIBOL_HOST_DEVICE void EvalWeakFormIntegral<COMMON_PLANE, SINGLE_POINT>( Surfac
                            elem.overlapNormal[2], cx[0], cx[1], cx[2], projX1[elem.dim * i], projX1[elem.dim * i + 1],
                            projX1[elem.dim * i + 2] );
 
+#ifdef TRIBOL_USE_HOST
```
srwopschall, line 82: Just confirm whether this should also check to see if tribol_debug is defined. You added TRIBOL_DEBUG, but not all SLIC_DEBUG statements are guarded.

srwopschall (approved):
@ebchin - I like the changes to the logging level enum. You added TRIBOL_DEBUG, but I only see one location where a SLIC_DEBUG_IF statement is guarded by checking TRIBOL_DEBUG. Many other locations just check TRIBOL_USE_HOST to guard a SLIC_DEBUG. Can you clarify?

CMakeLists.txt
ebchin: Nevermind, probably not needed!

CMakeLists.txt
ebchin: Correct

src/tribol/integ/FE.cpp
ebchin: `SLIC_DEBUG()` macros are contingent on the def of `AXOM_DEBUG`; if we want it contingent on `TRIBOL_DEBUG`, we would need to define our own macros.  `TRIBOL_USE_HOST` is to prevent SLIC calls on device.

src/tribol/integ/Integration.cpp
ebchin: Yeah, a bit confusing, but it holds.  See prev comment.

src/tribol/geom/GeomUtilities.cpp
```diff
@@ -1157,8 +1157,8 @@ TRIBOL_HOST_DEVICE bool SegmentIntersection2D( RealT xA1, RealT yA1, RealT xB1,
     xDiff = ( xDiff < 0. ) ? -1.0 * xDiff : xDiff;
     yDiff = ( yDiff < 0. ) ? -1.0 * yDiff : yDiff;
 
+#if defined( TRIBOL_DEBUG ) && defined( TRIBOL_USE_HOST ) && !defined( TRIBOL_USE_ENZYME )
```
srwopschall, line 1160: Can you clarify why we check to see if TRIBOL_DEBUG is defined here in addition to TRIBOL_USE_HOST etc., guarding this SLIC_DEBUG_IF() where other locations you just guard it with a #ifdef TRIBOL_USE_HOST? I am not clear on this yet.

chapman39 (approved):

src/tribol/geom/GeomUtilities.cpp
ebchin: Sure, I'll do my best:

- `TRIBOL_USE_HOST` is to make sure this function doesn't run on device, where SLIC is not compatible
- `TRIBOL_USE_ENZYME` is to make sure this function doesn't run on Enzyme, since axom is not built with `LLDEnzyme`
- `TRIBOL_DEBUG` is so the block of code only runs when `SLIC_DEBUG()` is active, preventing the otherwise unused variable `diffTol` from being defined

PR #144, opened 2025-04-16
Allow wider range of rate percent penalties

ebchin:
Relax the requirements on percent rate penalties to be between $[0, \infty)$.


srwopschall (approved):
@ebchin - good point for experimentation

PR #158, opened 2025-09-10
Tangential Viscous Damping

srwopschall:
This PR adds a tangential velocity based damping force term to the contact nodal response. This is taken as some damping coefficient scaling the tangential velocity gap vector. This is enabled using VISCOUS_TANGENTIAL as the contact model, and a damping coefficient for each mesh can be registered. The force is now added to the response arrays in ApplyTangential<>() for the common plane method only. A 2D and a 3D test were added to confirm that the normal component of any velocity gap vector is subtracted out and only the tangential component of the velocity gap is applied.

### DIFF
--- a/src/tests/tribol_common_plane_penalty.cpp
+++ b/src/tests/tribol_common_plane_penalty.cpp
@@ -594,6 +594,194 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
   compareGaps( couplingScheme, gap, 1.E-8, "kinematic_penetration" );
 }
 
+TEST_F( CommonPlaneTest, common_plane_viscous_tangential_2d )
+{
+  // this test has two edges in separation so there should be no contribution
+  // from the kinematic constraint enforcement, only the velocity based viscous
+  // tangential term
+  constexpr int numVerts = 2;
+
+  RealT x1[numVerts];
+  RealT y1[numVerts];
+  RealT x2[numVerts];
+  RealT y2[numVerts];
+
+  x1[0] = 1.0;
+  x1[1] = 0.0;
+  y1[0] = 0.0;
+  y1[1] = 0.0;
+
+  x2[0] = 0.0;
+  x2[1] = 1.0; 
+  y2[0] = 0.1; 
+  y2[1] = 0.1; 
+
+  tribol::IndexT conn1[numVerts] = { 0, 1 };
+  tribol::IndexT conn2[numVerts] = { 0, 1 };
+
+  tribol::registerMesh( 0, 1, numVerts, &conn1[0], (int)( tribol::LINEAR_EDGE ), &x1[0], &y1[0], nullptr,
+                        tribol::MemorySpace::Host );
+  tribol::registerMesh( 1, 1, numVerts, &conn2[0], (int)( tribol::LINEAR_EDGE ), &x2[0], &y2[0], nullptr,
+                        tribol::MemorySpace::Host );
+
+  RealT fx1[numVerts] = { 0., 0. };
+  RealT fy1[numVerts] = { 0., 0. };
+  RealT fx2[numVerts] = { 0., 0. };
+  RealT fy2[numVerts] = { 0., 0. };
+
+  // set nodal velocities in <1,1> and <-1,-1> directions
+  RealT vx1[numVerts] = { 1., 1. };
+  RealT vy1[numVerts] = { 1., 1. };
+  RealT vx2[numVerts] = { -1., -1. };
+  RealT vy2[numVerts] = { -1., -1. };
+
+  tribol::registerNodalResponse( 0, &fx1[0], &fy1[0], nullptr );
+  tribol::registerNodalResponse( 1, &fx2[0], &fy2[0], nullptr );
+
+  tribol::registerNodalVelocities( 0, &vx1[0], &vy1[0], nullptr );
+  tribol::registerNodalVelocities( 1, &vx2[0], &vy2[0], nullptr );
+
+  tribol::setKinematicConstantPenalty( 0, 1. );
+  tribol::setKinematicConstantPenalty( 1, 1. );
+
+  tribol::registerCouplingScheme( 0, 0, 1, tribol::SURFACE_TO_SURFACE, tribol::NO_CASE, tribol::COMMON_PLANE,
+                                  tribol::VISCOUS_TANGENTIAL, tribol::PENALTY, tribol::BINNING_GRID,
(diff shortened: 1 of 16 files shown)


srwopschall:
I have to add the viscous damping coefficients to mfem data. Right now updateMfemParallelRedecomposition() is not reassigning the damping coefficients, so any host code using the mfem api won't see tangential damping force contributions.

srwopschall:
@ebchin This is ready for review. The viscous damping feature has been added to the mfem api and tested elsewhere.

src/tribol/physics/Physics.hpp
```diff
@@ -11,8 +11,8 @@
 namespace tribol {
 
 // forward declarations
-struct InterfacePair;
-struct SurfaceContactElem;
+// struct InterfacePair;
+// struct SurfaceContactElem;
```
ebchin: Can these just be removed?  Are they not used anywhere in the header?

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -594,6 +594,206 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
...
+    EXPECT_NEAR( fx1[i], -.5, 1.e-10 );
+    EXPECT_NEAR( fy1[i], -force_y, 1.e-10 );
+    EXPECT_NEAR( fx2[i], .5, 1.e-10 );
+    EXPECT_NEAR( fy2[i], force_y, 1.e-10 );
+  }
+}
+
+TEST_F( CommonPlaneTest, common_plane_viscous_tangential_3d )
+{
+  // this test has two faces with zero gap so there should be no contribution
```
ebchin: Is the gap zero?  I see 0.01 on L712

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -594,6 +594,206 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
...
+  RealT fy2[numVerts] = { 0., 0., 0., 0. };
+  RealT fz2[numVerts] = { 0., 0., 0., 0. };
+
+  // set nodal velocities in <2,2,2> and <-2,-2,-2> directions
+  RealT vx1[numVerts] = { 2., 2., 2., 2. };
+  RealT vy1[numVerts] = { 2., 2., 2., 2. };
+  RealT vz1[numVerts] = { 2., 2., 2., 2. };
+  RealT vx2[numVerts] = { -2., -2., -2., -2. };
+  RealT vy2[numVerts] = { -2., -2., -2., -2. };
+  RealT vz2[numVerts] = { 2., 2., 2., 2. };
```
ebchin: This doesn't match the comment above.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -594,6 +594,206 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
   compareGaps( couplingScheme, gap, 1.E-8, "kinematic_penetration" );
 }
 
+TEST_F( CommonPlaneTest, common_plane_viscous_tangential_2d )
+{
+  // this test has two edges with zero gap so there should be no contribution
```
ebchin: Is the gap zero?  See L614.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -594,6 +594,206 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
...
+  EXPECT_EQ( update_err, 0 );
+
+  tribol::CouplingSchemeManager& couplingSchemeManager = tribol::CouplingSchemeManager::getInstance();
+
+  tribol::CouplingScheme* couplingScheme = &couplingSchemeManager.at( 0 );
+
+  EXPECT_EQ( 1, couplingScheme->getNumActivePairs() );
+
+  // check forces
+  RealT force_y = 0.5 * gap / numVerts;
```
ebchin, line 672: Where does this come from?  Shouldn't the visc_coeff be a part of it?  Why is the gap a part of it?  Isn't the viscosity only dependent on the tangential relative velocity?

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -594,6 +594,206 @@ TEST_F( CommonPlaneTest, common_plane_2d_interpen_check )
...
+  EXPECT_EQ( update_err, 0 );
+
+  tribol::CouplingSchemeManager& couplingSchemeManager = tribol::CouplingSchemeManager::getInstance();
+
+  tribol::CouplingScheme* couplingScheme = &couplingSchemeManager.at( 0 );
+
+  EXPECT_EQ( 1, couplingScheme->getNumActivePairs() );
+
+  // check forces
+  RealT force_z = 0.5 * gap / numVerts;
```
ebchin, line 787: Also clarify this force and where it comes from.  Should the x and y forces be zero or has the test been updated?

ebchin (commented):
The enforcement method code looks fine, just verify the test matches the comments and the expected force.

I didn't look very thoroughly, so I don't know if pieces of code are missing elsewhere (allocations/deallocations/mfem interface specifics), but I am OK merging after addressing the rest of the comments.

src/tribol/physics/Physics.hpp
srwopschall: These are now removed.

src/tests/tribol_common_plane_penalty.cpp
srwopschall: Fixed.

src/tests/tribol_common_plane_penalty.cpp
srwopschall: It was, but I updated it to have interpen. Updated comments.

src/tests/tribol_common_plane_penalty.cpp
srwopschall: There are tangential viscous forces in the x and y, and this is the calc for the spring force in the normal z-direction. Will add comments to clarify.

src/tests/tribol_common_plane_penalty.cpp
srwopschall: This is the normal force associated with two springs in parallel and a symmetric force distributed to all of the nodes. I added a comment for clarity.

src/tests/tribol_common_plane_penalty.cpp
srwopschall: Fixed.

ebchin (approved):

PR #162, opened 2025-09-24
Change tribol to use C++17

ebchin:


scripts/spack/packages/tribol/package.py
```diff
@@ -203,8 +203,6 @@ def initconfig_compiler_entries(self):
         else:
             entries.append(cmake_cache_option("ENABLE_FORTRAN", False))
 
-        entries.append(cmake_cache_string("BLT_CXX_STD", "c++14", ""))
```
srwopschall, line 206: is there a c++17 equivalent that should replace this line?

srwopschall (approved):

PR #165, opened 2025-10-02
Get Number of Contact Pairs

srwopschall:
Adding API functions to return the number of contact pairs per rank and total pairs across all ranks. This will help catch bugs/changes in the computational geometry filtering and cg algorithm in general.

This routine is host callable meant for diagnostics/metrics.

### DIFF
--- a/src/tribol/interface/tribol.hpp
+++ b/src/tribol/interface/tribol.hpp
@@ -498,7 +498,7 @@ void setInterfacePairs( IndexT cs_id, IndexT numPairs, IndexT const* mesh_id1, I
  *
  * \param [in] cs_id coupling scheme id
  *
- * \return the number of contact pairs on rank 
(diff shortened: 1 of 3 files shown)


src/tribol/interface/tribol.hpp
```diff
@@ -493,6 +493,26 @@ void setInterfacePairs( IndexT cs_id, IndexT numPairs, IndexT const* mesh_id1, I
...
+ *
+ * \param [in] cs_id coupling scheme id
+ *
+ * \return the number of contact pairs on rank
+ */
+int getNumberOfContactPairsOnRank( IndexT cs_id );
+
+/*!
+ * \brief Get the total number of contact pairs across all ranks for the given coupling scheme
+ *
```
ebchin, line 507: Add a warning re: ghost elements being counted multiple times.

PR #166, opened 2025-10-08
Small refactors to get debug build working with enzyme

ebchin:
The debug build was not compiling with enzyme due to errors like this:
```
PHI Node must not have any attached DbgRecords
  %53 = phi i64 [ %15, %19 ], !dbg !71744
; Function Attrs: mustprogress noinline optnone willreturn uwtable
```

This PR contains small refactors to get the code building with enzyme again on debug.


ebchin:
Rolled into #169 so CI will pass.

PR #171, opened 2025-10-11
Fix `my_plane` going out of scope

ebchin:


srwopschall (approved):
@ebchin - let's briefly discuss prior to merging, but this looks good and is quite a bit clearer/cleaner.

PR #172, opened 2025-10-13
Fix Gitlab CI

white238:
plus some minor clean up.


.gitlab/build_toss4.yml
```diff
@@ -17,19 +17,16 @@
   rules:
     - if: '$CI_COMMIT_BRANCH =~ /_qnone/ || $ON_TOSS4 == "OFF"' #run except if ...
       when: never
+    - when: on_success
```
white238, line 20: This was the fix. It different because of something to do with the base rules that Smith has at the root file.

ebchin (approved):
Thanks @white238!

srwopschall (approved):

PR #179, opened 2025-11-18
Add Caliper to spack recipe and CMake

ebchin:
Adds Caliper as a TPL of Tribol for profiling Tribol routines.  No annotations have been added in this PR, just the plumbing to use Caliper in Tribol and in downstream packages.

Specifically, the following has been added:
- CMake build system changes required for building with Caliper support
- Spack recipe changes for building Caliper with TPLs when `+profiling` variant is active
- Tribol macros for more fine-tuned profiling output
- Caliper smoke test

NOTE: All of this was heavily borrowed from [Smith](https://github.com/LLNL/smith).  Thanks @chapman39 and @white238!

### DIFF
--- a/src/tribol/utils/Profiling.hpp
+++ b/src/tribol/utils/Profiling.hpp
@@ -1,0 +1,79 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Tribol Project Developers. See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: (MIT)
+
+#ifndef SRC_TRIBOL_UTILS_PROFILING_HPP_
+#define SRC_TRIBOL_UTILS_PROFILING_HPP_
+
+/**
+ * @file profiling.hpp
+ *
+ * @brief Various helper functions and macros for profiling using Caliper
+ */
+
+// Tribol config include
+#include "tribol/config.hpp"
+
+#ifdef TRIBOL_USE_CALIPER
+#include "caliper/cali.h"
+#endif
+
+/**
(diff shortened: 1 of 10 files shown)


chapman39 (approved):

src/tribol/utils/Profiling.hpp
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2017-2025, Lawrence Livermore National Security, LLC and
+// other Tribol Project Developers. See the top-level LICENSE file for details.
+//
```
srwopschall, line 3: Do we want to put this in shared for any reason?

src/tribol/utils/Profiling.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+// Tribol config include
+#include "tribol/config.hpp"
+
+#ifdef TRIBOL_USE_CALIPER
+#include "caliper/cali.h"
+#endif
+
+/**
+ * @def TRIBOL_MARK_FUNCTION
```
srwopschall, line 23: Just so I know. Was this stuff brought over from another project?

srwopschall (approved):

src/tribol/utils/Profiling.hpp
chapman39: yes, smith

src/tribol/utils/Profiling.hpp
ebchin: I started adding caliper macros and it quickly became apparent I needed them in redecomp too.  I will do this in a follow on PR where I add the markings, if that's OK?

PR #180, opened 2025-11-19
Add caliper annotations to `updateMfemParallelDecomposition()`

ebchin:
This PR does 2 things:
- moves caliper to `tribol_shared` to allow profiling in redecomp and Tribol
- adds annotations to `updateMfemParallelDecompsition()` and the functions that it calls to get a better idea of where time is spent in this routine


srwopschall (approved):
Looks good. Just double check that you've added annotations where appropriate and/or is useful.

Issue #174, opened 2025-11-06
DevOps Tasks


chapman39:
tasks to help tribol's devops systems

- [x] convert from azure to github actions https://github.com/LLNL/Tribol/pull/181
- [x] update gitlab ci, python build scripts, etc.
- [x] add docs/ doxygen ci check https://github.com/llnl/Tribol/pull/183
- [x] test tribol on macmini https://github.com/llnl/Tribol/pull/185
- [ ] add cuda docker containers
- [ ] tribol install test
- [ ] TPL build CI job
- [ ] (maybe) adding benchmarking infrastructure, creating one simple benchmark for reference

on-going tasks that should be done as a regular basis

* tpl updates, moving and converting smith host configs to tribol
* other general ci/ cmake/ python script developments

@white238 for awareness

chapman39:
i created individual issues tracking the rest of the tasks, so i will close this one for now.

PR #181, opened 2025-11-25
Switch docker container testing to use Github Actions instead of Azure

chapman39:
- [x] Migrates Azure Pipelines to GitHub Actions
- [x] Updates `scripts/llnl` build scripts
- [x] Updates docker host configs

Some benefits of GitHub Actions:

* You can now cancel a job at any point.
* More storage (14GB instead of Azure's 10GB) (sources: [azure](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#capabilities-and-limitations), [github actions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)).
* Jobs will no longer run on push. They will only run if there is an update to a pull request.
* Previous of a PR jobs are automatically canceled if an update has been made.
* Fails fast. If one job fails, the entire pipeline will cancel.
* You can re-try failed jobs.

To do before merging:

* Make GitHub Actions required to pass before merging into develop https://stackoverflow.com/a/60779334
* Remove Azure pipeline


chapman39:
only things thats left to do is make github actions required and removing azure from the branch protection rules. once this PR is approved, i'll update those settings and merge shortly after

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,124 @@
...
+    strategy:
+      fail-fast: true
+      matrix:
+        build_type: [ Debug, Release ]
+        config:
+        - job_name: llvm@19.1.1, shared
+          host_config: llvm@19.1.1.cmake
+          compiler_image: ${{ needs.set_image_vars.outputs.clang_docker_image }}
+          cmake_opts: "-DENABLE_DOCS=OFF -DBUILD_REDECOMP=ON"
+        # TODO re-add cuda testing
```
ebchin, line 43: No need to tackle this in this PR, but of curiosity, did you try this out?

scripts/llnl/common_build_functions.py
```diff
@@ -268,95 +280,101 @@ def build_and_test_host_config(test_root, host_config, report_to_stdout=False, e
...
-    #     return res
+    if res != 0:
+        print("[ERROR: Docs generation for host-config: %s failed]\n\n" % host_config)
+        return res
+    end_time = time.time()
+    print("[docs time: {0}]\n".format(convertSecondsToReadableTime(end_time - start_time)))
 
     # Install and test examples
-    # TODO: skipped for now until Tribol has a install test
+    # TODO add tribol install test
```
ebchin, line 358: This seems like a good idea to add.  I see it on #174 with a question mark, but it might make sense to move it up the list?

ebchin (approved):
Thanks @chapman39!  Very excited to use this.

scripts/llnl/common_build_functions.py
chapman39: that sounds good to me. similar to the other tasks, i can use smith as a reference, so shouldn't be too much work. ill remove that question mark, lol.

.github/workflows/ci-tests.yml
chapman39: i didn't try it. when i updated the docker host configs, i took the smith ones and modified them only for what tribol needs. but cuda testing is disabled in smith, so there's no cuda host config.

i would like to revisit this sometime in the near future.

srwopschall (approved):
@chapman39 - Thanks for putting this together. This is definitely an enhancement.

PR #182, opened 2025-12-04
Use mfem enzyme header

chapman39:
remove tribol enzyme header, since its more-or-less the same as the mfem one.

depends on https://github.com/LLNL/Tribol/pull/183


ebchin (approved):
Thanks @chapman39!

PR #188, opened 2026-01-17
GPU aware transfers in Redecomp

ebchin:
This PR adds GPU aware transfer operators for redecomp.


scripts/spack/packages/mfem/package.py
```diff
@@ -11,7 +11,7 @@ class Mfem(BuiltinMfem):
     # Note: Make sure this sha coincides with the git submodule
     # Note: We add a number to the end of the real version number to indicate that we have
     # moved forward past the release. Increment the last number when updating the commit sha.
-    version("4.9.0", commit="d9d6526cc1749980a2ba1da16e2c1ca1e07d82ec")
+    version("4.9.0.2", commit="72805e340996a73179082bbe1c7bfb785c2afb47")
```
srwopschall, line 14: Is this consistent with tpl updates etc.?

scripts/spack/specs.json
```diff
@@ -28,6 +28,7 @@
 
     "linux_ubuntu_24":
     [ "+raja+umpire+enzyme+profiling ^hypre+int64 %clang_19",
-      "+cuda+raja+umpire+enzyme+profiling cuda_arch=86 ^hypre+int64 %clang_19" ]
+      "+cuda+raja+umpire+enzyme cuda_arch=89 ^openmpi+cuda cuda_arch=89 fabrics=ucx ^ucx+cuda cuda_arch=89 ^hypre+int64 %clang_19"
```
srwopschall, line 31: I'm not familiar with this line but cuda_arch=89 is in there twice and there was some openmpi stuff omitted elsewhere. Please verify.

src/redecomp/common/CoordList.hpp
```diff
@@ -0,0 +1,66 @@
...
+namespace redecomp {
+
+/**
+ * @brief CoordList class that holds a 2D mfem::Vector and provides access to points.
+ *
+ * @tparam NDIMS Number of dimensions
+ * @tparam ORDERING Element DOF ordering (LEXICOGRAPHIC or NATIVE)
+ */
+template <int NDIMS>
+class CoordList {
```
srwopschall, line 21: Is there functionality to grab the (x,y,z) coordinates for a single node? Is that through the GetPoint() and the Point class?

src/redecomp/common/CoordList.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * @brief Get the underlying coordinate vector
+   *
+   * @return const mfem::Vector&
+   */
+  const mfem::Vector& GetCoords() const { return coords_; }
+
+  /**
+   * @brief Get the number of coordinates
```
srwopschall, line 54: Is this the same as number of nodes?

src/redecomp/partition/PartitionElements.cpp
```diff
@@ -5,21 +5,65 @@
...
+
 namespace redecomp {
 
 template <int NDIMS>
-std::vector<axom::Array<Point<NDIMS>>> PartitionElements<NDIMS>::EntityCoordinates(
+std::vector<CoordList<NDIMS>> PartitionElements<NDIMS>::EntityCoordinates(
     const std::vector<const mfem::ParMesh*>& par_meshes ) const
 {
-  auto elem_centroids = std::vector<axom::Array<Point<NDIMS>>>();
+  // check if device is available to MFEM
```
srwopschall: This comment seems out of place.

src/redecomp/partition/PartitionElements.cpp
```diff
@@ -5,21 +5,65 @@
...
+
   for ( auto par_mesh : par_meshes ) {
     auto n_elems = par_mesh->GetNE();
-    elem_centroids.emplace_back( n_elems, n_elems );
-    for ( int i{ 0 }; i < n_elems; ++i ) {
-      auto vec_centroid = mfem::Vector( elem_centroids.back()[i].data(), NDIMS );
-      // TODO: const version of GetElementCenter()
-      const_cast<mfem::ParMesh*>( par_mesh )->GetElementCenter( i, vec_centroid );
+    if ( n_elems == 0 ) {
+      elem_centroids.emplace_back( mfem::Vector() );
```
srwopschall, line 22: What does this line 23 do? If there are zero elements, is this just a null coord list?

src/redecomp/partition/RCB.cpp
```diff
@@ -37,31 +38,35 @@ std::vector<EntityIndexByRank> RCB<NDIMS>::generatePartitioning(
...
-        if ( problem_tree( neighbors[j] ).ghost_bbox_.contains( coords[i] ) ) {
+        if ( problem_tree( neighbors[j] ).ghost_bbox_.contains( coord ) ) {
           auto neighbor_rank = neighbors[j] * rank_jump_interval;
           ent_idx[neighbor_rank].push_back( i );
           ent_ghost[neighbor_rank].push_back( true );
         }
       }
     }
     for ( int i{ 0 }; i < n_parts; ++i ) {
-      ent_idx[i].shrink();
```
srwopschall, line 63: Can you clarify this change? I imagine functionally its the same thing?

src/redecomp/partition/RCB.cpp
```diff
@@ -231,13 +236,15 @@ BisecTree<RCBInfo<NDIMS>> RCB<NDIMS>::BuildProblemTree( int n_parts,
...
-BoundingBox<NDIMS> RCB<NDIMS>::DomainBoundingBox( const std::vector<axom::Array<Point<NDIMS>>>& coords_by_mesh ) const
+BoundingBox<NDIMS> RCB<NDIMS>::DomainBoundingBox( const std::vector<CoordList<NDIMS>>& coords_by_mesh ) const
 {
   auto on_rank_bbox = BoundingBox<NDIMS>();
   size_t num_coords{ 0 };
   for ( const auto& coords : coords_by_mesh ) {
-    on_rank_bbox.addBox( BoundingBox<NDIMS>( coords.data(), coords.size() ) );
-    num_coords += coords.size();
+    for ( int i = 0; i < coords.GetNumCoords(); ++i ) {
+      on_rank_bbox.addPoint( coords.GetPoint( i ) );
```
srwopschall, line 245: It looks like the bounding box is created by adding points now and not by adding a box? This doesn't change anything does it?

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -222,7 +222,8 @@ MPIArray<int> MatrixTransfer::buildSendArrayIDs( const axom::Array<int>& test_el
     }
   }
   for ( auto send_array_ids_rank : send_array_ids ) {
-    send_array_ids_rank.shrink();
+    auto tmp_send_array_ids_rank = send_array_ids_rank;
+    std::swap( send_array_ids_rank, tmp_send_array_ids_rank );
```
srwopschall: Verify this swap is correct. Elsewhere where there was a call to shrink() it was replaced by a std::move.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -323,13 +324,13 @@ MPIArray<HYPRE_BigInt> MatrixTransfer::buildRecvTrialElemDofs( const RedecompMes
   // List of trial element offsets sorted by the parent test space rank and the
   // parent trial space rank it belongs to.  Used to get the trial space parent
   // vdofs (on the parent trial space rank) onto the parent test space rank.
-  auto send_trial_elem_offsets = MPIArray<axom::Array<int>>( &getMPIUtility() );
+  auto send_trial_elem_offsets = MPIArray<mfem::Array<int>, std::vector<mfem::Array<int>>>( &getMPIUtility() );
```
srwopschall, line 327: Can you clarify the change from one template arg to two?

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -381,40 +381,40 @@ MPIArray<HYPRE_BigInt> MatrixTransfer::buildRecvTrialElemDofs( const RedecompMes
...
+        [&recv_trial_elem_offsets]( mfem::Array<int>&& send_vals, axom::IndexType src_trial_r ) {
           recv_trial_elem_offsets[src_trial_r] = std::move( send_vals );
         } );
     // send parent trial vdofs back to test redecomp rank
     auto first_dof = parent_trial_fes_.GetMyDofOffset();
     auto trial_dofs_by_rank = MPIArray<HYPRE_BigInt>( &getMPIUtility() );
     trial_dofs_by_rank.SendRecvEach(
         [this, &trial_p2r_elems, &recv_trial_elem_offsets, first_dof]( axom::IndexType src_trial_r ) {
-          auto trial_dofs = axom::Array<HYPRE_BigInt>();
+          auto trial_dofs = MPIArray<HYPRE_BigInt>::ArrayT();
```
srwopschall, line 405: I think this is all fine, but it is confusing as to why you go from axom::Array to mfem::Array or to std::vector, or here to MPIArray. I'm not seeing the rationale, but if it works it works.

src/redecomp/utils/MPIArray.hpp
```diff
@@ -42,79 +40,83 @@ class MPIArray : public ArrayType {
    * @param mpi MPIUtility to define MPI_Comm for MPI operations
    * @param array Array data
    */
-  MPIArray( const MPIUtility* mpi, ArrayType&& array ) : ArrayType( std::move( array ) ), mpi_{ mpi }
+  MPIArray( const MPIUtility* mpi, std::vector<ArrayType>&& array )
+      : std::vector<ArrayType>( std::move( array ) ), mpi_{ mpi }
   {
-    this->reserve( mpi_->NRanks() );
```
srwopschall, line 47: Confirm we don't need these three lines anymore.

src/redecomp/utils/MPIArray.hpp
```diff
@@ -15,25 +15,23 @@
...
   /**
    * @brief Construct a new MPIArray object
    *
    * @param mpi MPIUtility to define MPI_Comm for MPI operations
    * @param array Array data
    */
-  MPIArray( const MPIUtility* mpi, const ArrayType& array ) : ArrayType( array ), mpi_{ mpi }
+  MPIArray( const MPIUtility* mpi, const std::vector<ArrayType>& array ) : std::vector<ArrayType>( array ), mpi_{ mpi }
   {
-    this->reserve( mpi_->NRanks() );
```
srwopschall, line 34: Confirm we don't need these three lines anymore. Is the constructor a no-op outside the initializer?

src/redecomp/utils/MPIUtility.cpp
```diff
@@ -66,4 +49,65 @@ MPI_Datatype MPIUtility::GetMPIType<long>() const
...
+      auto right_node_it = send_tree.right( lvl_it, node_it );
+      *node_it = right_node_it.second == send_tree.end( right_node_it.first ) || *right_node_it.second != rank
+                     ? *send_tree.left( lvl_it, node_it ).second
+                     : rank;
+    }
+  }
+  return send_tree;
+}
+
+}  // end namespace redecomp
```
srwopschall: make sure to run formatting.

src/redecomp/utils/MPIUtility.hpp
```diff
@@ -344,8 +440,29 @@ T MPIUtility::Recv( type<T>, int source, int tag ) const
...
+  } else {
+    data = container.HostWrite();
+  }
+  MPI_Recv( data, count, GetMPIType<typename std::remove_cv<T>::type>(), source, tag, comm_, &status_ );
+  return container;
+}
+
 template <typename T, axom::MemorySpace Sp>
-axom::Array<T, 2, Sp> MPIUtility::Recv( type<axom::Array<T, 2, Sp>>, int source, int tag ) const
+axom::Array<T, 2, Sp> MPIUtility::Recv( type<axom::Array<T, 2, Sp>>, int source, int tag, bool ) const
```
srwopschall, line 465: Confirm adding just "bool" at the end here.

src/redecomp/RedecompTransfer.cpp
```diff
@@ -112,26 +110,24 @@ void RedecompTransfer::TransferToParallel( const mfem::QuadratureFunction& src,
...
         // skip ghost elements
-        if ( ghost_ct < redecomp->getRedecompToParentGhostElems()[dest].size() &&
+        if ( ghost_ct < redecomp->getRedecompToParentGhostElems()[dest].Size() &&
              redecomp->getRedecompToParentGhostElems()[dest][ghost_ct] == e ) {
           ++ghost_ct;
         } else {
           src.GetValues( e, vals );
-          src_vals.insert( quadpt_ct, vals.Size(), vals.GetData() );
-          quadpt_ct += vals.Size();
+          src_vals.Append( vals.GetData(), vals.Size() );
```
srwopschall, line 130: This looks simpler and not having to carry around the quadpt_ct.

src/tests/tribol_mfem_mortar_lm.cpp
```diff
@@ -146,7 +145,6 @@ class MfemMortarTest : public testing::TestWithParam<std::tuple<int, mfem::Eleme
                                         tribol::BINNING_GRID );
     tribol::setLagrangeMultiplierOptions( 0, tribol::ImplicitEvalMode::MORTAR_RESIDUAL_JACOBIAN );
 
-    coords.ReadWrite();
```
srwopschall, line 149: Verify removing these coords.ReadWrite() calls.

src/tribol/mesh/MfemData.cpp
```diff
@@ -33,17 +34,13 @@ void SubmeshLORTransfer::TransferToLORGridFn( const mfem::ParGridFunction& subme
 void SubmeshLORTransfer::TransferFromLORVector( mfem::Vector& submesh_dst ) const
 {
   // make sure host data is up to date.  this transfer needs to be on the host until submesh supports device transfer
-  lor_gridfn_->HostRead();
```
srwopschall, line 38: Clarify removing these HostRead() and HostWrite() calls.

src/tribol/mesh/MfemData.cpp
```diff
@@ -53,9 +50,6 @@ std::unique_ptr<mfem::ParGridFunction> SubmeshLORTransfer::CreateLORGridFunction
   auto lor_gridfn = std::make_unique<mfem::ParGridFunction>(
       new mfem::ParFiniteElementSpace( &lor_mesh, lor_fec.get(), vdim, mfem::Ordering::byNODES ) );
   lor_gridfn->MakeOwner( lor_fec.release() );
-  // NOTE: This needs to be false until submesh supports device transfer. Otherwise, there will be extra copies to/from
-  // device.
-  lor_gridfn->UseDevice( false );
```
srwopschall, line 60: Is whether device is being used set elsewhere now?

srwopschall (approved):
Looks good. I could not do an in depth review of transfer by elements and transfer by nodes just cause of the complexity. I have some clarify/verify comments. Thanks @ebchin - Make sure everything is up to date to pass CI.

scripts/spack/packages/mfem/package.py
ebchin: Just updated this MFEM version to one that has the fixes in it needed. It's not consistent with TPLs anymore, Smith has moved forward. Once the fixes are in an MFEM version, we should update our package.py to note the required version of MFEM.

scripts/spack/specs.json
ebchin: Looks OK

src/redecomp/common/CoordList.hpp
ebchin: Yes

src/redecomp/common/CoordList.hpp
ebchin: No, the `CoordList` doesn't necessarily have all of the nodes.

src/redecomp/partition/PartitionElements.cpp
ebchin: Good catch, removed.

src/redecomp/partition/PartitionElements.cpp
ebchin: Correct

src/redecomp/partition/RCB.cpp
ebchin: Yeah, same thing. Just doing the manual copy to a appropriately sized vector since there isn't a `shrink()` method.

src/redecomp/partition/RCB.cpp
ebchin: They do the same thing, since we don't have a vector of `Point`s available anymore, we can't use `addBox()`.

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: Thanks for pointing this out. Changed it to `std::move`.

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: The second (optional) arg lets you specify the type of the on-rank array.  It looks like I decided to make this a `std::vector` instead of an `mfem::Array`, probably to make the `resize()` member available?

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: This particular change is meant to prevent bugs, since it forces `trial_dofs` to use the per-rank array type of the `MPIArray`.

More generally, I tried to use the container most well-suited to the requirements of the data being stored (interoperability with TPLs, dimension of the array, etc.).

src/redecomp/utils/MPIArray.hpp
ebchin: The three lines are there to force the (what was previously an) `axom::Array` to have the right size.  `std::vector` should create the right sized array without these lines.

src/redecomp/utils/MPIArray.hpp
ebchin: Not needed

src/tests/tribol_mfem_mortar_lm.cpp
ebchin: Unless we do something explicitly on host/device in this code, we shouldn't need `ReadWrite()` etc. in these test drivers.  This may have been to explicitly test what happens when data is on device, but this probably isn't needed anymore (Tribol should copy things to device when required).

src/tribol/mesh/MfemData.cpp
ebchin: Looks OK.  All of the needed calls to `HostRead()` etc. are in the transfer operator itself, and shouldn't be needed here.

src/tribol/mesh/MfemData.cpp
ebchin: Yeah, we are using the GPU-friendly partial assembly version of LOR where available now.

PR #192, opened 2026-01-22
Update tribol spack package

chapman39:
I diff'ed between Smith's tribol spack package and Tribol's tribol spack package, and took what was good from both of them. This is the result. Also changed the +profiling variant to +caliper, because that's the norm with Axom, Smith, etc.


ebchin (approved):
Looks good @chapman39

PR #196, opened 2026-02-14
Simplified Jacobian Transfers

ebchin:
- Introduces a new MatrixTransfer method for transferring a single (non-block) matrix from the redecomp mesh to finite element spaces defined on a parent mesh.
- Introduces a new GetMfemJacobian() that maps Tribol element Jacobian data to a parent mesh (calling the new MatrixTransfer method)
- Adds a test and example comparing timing for the old method (block matrices) and new method (single matrix only)


src/tribol/mesh/MfemData.hpp
```diff
@@ -1487,6 +1487,17 @@ class MfemSubmeshData {
   MfemSubmeshData( mfem::ParSubMesh& submesh, mfem::ParMesh* lor_mesh,
                    std::unique_ptr<mfem::FiniteElementCollection> pressure_fec, int pressure_vdim, bool use_device );
 
+  /**
+   * @brief Update (or clear) the LOR mesh used for pressure/gap transfers
```
srwopschall, line 1491: Even with the doxygen comments, it is unclear to me exactly what this routine is doing, except for "clear". Does this routine actually "set" the lor mesh and if so, perhaps include a high level comment about what "setting" means.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1487,6 +1487,17 @@ class MfemSubmeshData {
...
+   *
+   * @note This should be called if the owning MfemMeshData rebuilds its LOR mesh
+   * (e.g., via MfemMeshData::SetLORFactor()). This resets internal transfer data
```
srwopschall, line 1494: This comment doesn't seem accurate. Does `SetLORFactor()` actually build/re-build the LOR mesh?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1487,6 +1487,17 @@ class MfemSubmeshData {
...
+   * so the next UpdateMfemSubmeshData() rebuilds consistent redecomp transfers.
```
srwopschall, line 1495: Minor: "rebuilds redecomp transfers consistent with the updated/built LOR mesh" (?) (or something like that).

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
   bool use_device_;
 };
 
+/**
+ * @brief Solver-facing role of a variable space
```
srwopschall: Consider giving an example of a Primary vs. a Dual variable. I have not yet seen where/how this is used so maybe it is obvious, but I'm mixed up between primal, dual, and now primary, and if we are using lagrange multipliers or if we aren't etc. Consider clarifying.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ */
+enum class VariableRole
+{
+  Primary,
+  Dual
+};
+
+/**
+ * @brief Surface partition associated with a contribution source
```
srwopschall: Elaborate comment. What does a "surface partition" mean here? And what is a "contribution"? A "contribution" to what? 

Again, I haven't gotten to where this is used, but this isn't clear. Is `Mortar`, `Nonmortar`, and `Other` contribution sources? Or are they surface partitions? Or both? And what role does `Other` play? Should we have an `Undefined`?

Lastly, just verify code style with enums vs. enum classes as we used full capitalization elsewhere.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ */
+enum class SurfacePartition
+{
+  Mortar,
+  Nonmortar,
+  Other
+};
+
+/**
+ * @brief Combined description of a contribution's algebraic role and surface partition
```
srwopschall: Comment is a little vague, but I think once the member enum's comments are updated it will be fine.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ */
+struct ContributionSpace {
+  VariableRole role;
+  SurfacePartition partition;
+};
+
+/**
+ * @brief Base class for Jacobian-related operators that can also materialize an explicit matrix
```
srwopschall: "Materialize an explicit matrix" is a little confusing. Is this an assembly operator that wraps an mfem::Operator that can assemble an implicit contact method's Jacobian contributions or an local and/or global operator for explicit methods? It is vague as to what an "explicit matrix" is, although I think I know where you are going with this.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ */
+class AssembleableParOperator : public mfem::Operator {
```
srwopschall: "Assembleable" is a mouthful and hard to type haha. Can we just assume that "assembly" is a parallel thing and call this AssemblyOperator? Or MatrixAssemblyOperator? Or something like that?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ public:
+  /**
+   * @brief Construct an operator with the given local row and column sizes
```
srwopschall: Minor: "an operator" is vague. Try to make the comment a little more Tribol contact specific for added context?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  AssembleableParOperator( int height, int width );
+
+  /**
+   * @brief Apply the operator to a vector
+   *
+   * @note The default implementation assembles and caches the matrix, then applies it.
```
srwopschall: I'm sure I'll get there, but at this point it is unclear how or where the matrix is assembled. The constructor on line 1694 looks like it just sizes the operator. Consider a note that if another routine is to be called prior to Mult() to actually assemble the operator's contributions then what is that call order?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  void Mult( const mfem::Vector& x, mfem::Vector& y ) const override;
```
srwopschall, line 1683: Consider comment of what x and y are. Is y the Operator * x return Vector? Consider \param [in] etc. for consistency with other parts of the code.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+
+  /**
+   * @brief Materialize the operator as an explicit parallel sparse matrix
```
srwopschall: "Materialize" sounds mystical. Consider wording. Also, what is the difference or delineation between an explicit matrix and/or implicit matrix. If they are different, consider naming this routine `AssembleExplicit()`. If they are the same, then can we just drop the word explicit and just call an operator an operator and the developer has to be in charge of what that means for their method. Or does "explicit" mean something entirely different? If so, it is an overloaded word.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  virtual shared::ParSparseMat Assemble() const = 0;
+
+  /**
+   * @brief Return a cached assembled matrix, assembling it on first use
```
srwopschall: Clarify "assembling it on first use". Is this just a getter or does it actually assemble, and if so, does something need to be called first to make sure it has the data required to assemble?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  const shared::ParSparseMat* GetAssembled() const;
+
+ protected:
+  /**
+   * @brief Build the assembly cache if needed and return it by reference
```
srwopschall: Clarify "assembly cache". Is this just the operator that is built (if not already built) and then returned as a ParSparseMat? Consider possible duplication of functionality and clarity in carrying around a `GetAssembled()` and `GetOrAssembleCache()`.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  shared::ParSparseMat& GetOrAssembleCache() const;
+
+ private:
+  mutable std::unique_ptr<shared::ParSparseMat> assembled_cache_;
```
srwopschall: Add comment. Assembled Cache aren't two words thrown around together very often. If this is just the operator, can we just call it `operator` or `assembled_operator`? If we have a ParSparseMat operator isn't it understood that at some point it has to be assembled?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   * @brief Construct an operator that maps HO DOFs to LOR DOFs for one field space
+   */
+  HoToLorTransferOp( const mfem::ParFiniteElementSpace& ho_fes, const mfem::ParFiniteElementSpace& lor_fes,
+                     const mfem::ParFiniteElementSpace& ho_scalar_fes,
+                     const mfem::ParFiniteElementSpace& lor_scalar_fes );
+
+  /**
+   * @brief Apply MFEM's runtime HO->LOR transfer without explicitly assembling a matrix
+   */
+  void Mult( const mfem::Vector& x, mfem::Vector& y ) const override;
```
srwopschall: I have to look at the implementation, but `Mult()` is very general, while this particular HoToLorTransferOp is a very specific class. One consideration is to not make Mult() virtual and then have a `ApplyHoToLorTransfer()` routine, for instance, that wraps a Mult() call? This may make the code more readable? These are just design decisions. 

Is there a LOR->HO transfer class that also has a "Mult()" implementation?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+
+  /**
+   * @brief Assemble a sparse matrix matching MFEM's HO->LOR transfer action
+   */
+  shared::ParSparseMat Assemble() const override;
```
srwopschall: I see that this class can assemble a LOR transfer matrix, but there does not appear to be a function that does the HO->LOR transfer with that matrix? I only see the Mult() which states that it does not use the assembled matrix. Can you clarify?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+
+ private:
+  const mfem::ParFiniteElementSpace& ho_fes_;
+  const mfem::ParFiniteElementSpace& lor_fes_;
+  const mfem::ParFiniteElementSpace& ho_scalar_fes_;
+  const mfem::ParFiniteElementSpace& lor_scalar_fes_;
+  mutable std::unique_ptr<mfem::L2ProjectionGridTransfer> transfer_;
```
srwopschall: I'm assuming we always want the L2ProjectionGridTransfer? Also, does this class need an LOR factor?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+class SubmeshParentTransferOp : public AssembleableParOperator {
+ public:
+  /**
+   * @brief Construct an operator that injects boundary-submesh DOFs into the parent FE space
+   */
+  SubmeshParentTransferOp( const mfem::ParFiniteElementSpace& submesh_fes,
+                           const mfem::ParFiniteElementSpace& parent_fes,
+                           const mfem::Array<HYPRE_BigInt>& submesh2parent_vdof_list );
+
+  /**
```
srwopschall, line 1745: Clarifying question. There is no Mult() routine here. Does every derived class need a Mult() implementation? Isn't this the way to apply the operator?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+
+/**
+ * @brief Struct to hold computed element data for Jacobian assembly
+ */
+struct ComputedElementData {
+  BlockSpace row_space{ BlockSpace::MORTAR };  ///< Block space for row elements
+  BlockSpace col_space{ BlockSpace::MORTAR };  ///< Block space for column elements
+  axom::Array<int> row_elem_ids;               ///< Tribol element IDs for rows
+  axom::Array<int> col_elem_ids;               ///< Tribol element IDs for columns
+  axom::Array<double> jacobian_data;           ///< Flattened Jacobian data
```
srwopschall: Is jacobian_data 1D array data?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ private:
+  const mfem::ParFiniteElementSpace& submesh_fes_;
+  const mfem::ParFiniteElementSpace& parent_fes_;
+  const mfem::Array<HYPRE_BigInt>& submesh2parent_vdof_list_;
+};
+
+/**
+ * @brief Struct to hold computed element data for Jacobian assembly
+ */
+struct ComputedElementData {
```
srwopschall: Minor: "Computed" is vague. Element data could be residuals. Maybe just call "JacobianElementData"? 

Just to clarify, this looks like a helper struct that simply holds element jacobian data in a useful way?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+
+  ComputedElementData( BlockSpace row_space_in, BlockSpace col_space_in )
+      : row_space( row_space_in ), col_space( col_space_in )
+  {
+  }
+
+  /**
+   * @brief Reserve packed storage for a batch of element contributions
+   */
+  void reserve( int n_entries, int n_values )
```
srwopschall: Minor: Clarify what n_values is. Is this number of elements? Or is this the number of individual values stored in the flattened array across all elements?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+  }
+
+  /**
+   * @brief Number of element contributions stored in this packed block
+   */
+  int numEntries() const { return row_elem_ids.size(); }
+};
+
+/**
+ * @brief Jacobian stage operator from redecomp element contributions to a surface FE space pair
```
srwopschall: I think unless you wrote redecomp, no one will understand this comment. What is a "stage operator" and what does it do between a redecomp element contribution and a surface FE space pair? What is a surface FE space pair? Is that a face-face pair? Is a redecomp element one redecomp mesh face on one side of a contact interaction? Or is it a face-face contact element? Consider elaborating.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+ */
+class RedecompJacobianStageOp : public AssembleableParOperator {
+ public:
+  /**
+   * @brief Construct one redecomp->surface Jacobian stage for a logical row/column pairing
```
srwopschall: Clarify comment here: when I see redecomp stated I think a redecomp mesh, which is a surface mesh. So clarify what redecomp->surface means.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   */
+  RedecompJacobianStageOp( const redecomp::MatrixTransfer& transfer, std::vector<ComputedElementData> contributions,
```
srwopschall: Add doxygen for input arguments here and elsewhere.

It looked like the struct, ComputedElementData can hold data for many elements, but here we have a std::vector of ComputedElementData. From a design standpoint is this just for flexibility that you can compute, for instance, individual blocks and store them in a single ComputedElementData, but then carry around a std::vector of ComputedElementData-s for all blocks?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+                           const Array1D<int>& mortar_elem_map, const Array1D<int>& nonmortar_elem_map, int height,
```
srwopschall: I think this is ok, but was wondering if mortar/nonmortar is still appropriate if using a matrix/jacobian based method for explicit dynamics, but if we do this it will likely be a "mortar" method so this naming still is appropriate.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+                           int width );
+
+  /**
+   * @brief Assemble the redecomp-stage matrix from flattened element contributions
+   */
+  shared::ParSparseMat Assemble() const override;
+
+ private:
+  const redecomp::MatrixTransfer& transfer_;
```
srwopschall, line 1870: Doxygen on member variable.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+  std::vector<ComputedElementData> contributions_;
+  const Array1D<int>& mortar_elem_map_;
+  const Array1D<int>& nonmortar_elem_map_;
+};
+
+/**
+ * @brief Solver-visible logical Jacobian block assembled from one stage operator
```
srwopschall: You are going to have to clarify what this means. What is the intended use of this struct? What is a "logical Jacobian block"

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+        dual_ho_to_lor( dual_ho_to_lor_in ),
+        submesh_parent_xfer( submesh_parent_xfer_in ),
+        parent_fes( parent_fes_in ),
+        dual_fes( dual_fes_in ),
+        inactive_dual_tdofs( inactive_dual_tdofs_in )
+  {
+  }
+};
+
+class LogicalJacobianBlockOp : public AssembleableParOperator {
```
srwopschall: Does this class need to be derived off the AssembleableParOperator? It sounds like the intent is that this just constructs a block of data from a stage operator and transfer operator, which are obviously both derived off the AssembleableParOperator. Let's discuss.

Also, "Logical" doesn't mean anything to me here :)

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+  LogicalJacobianBlockOp( VariableRole row_role, VariableRole col_role,
+                          std::unique_ptr<RedecompJacobianStageOp> stage_op, LogicalBlockTransferContext context );
+
+  /**
+   * @brief Assemble the full logical block on solver-facing true-dof spaces
+   */
+  shared::ParSparseMat Assemble() const override;
+
+ private:
+  const HoToLorTransferOp* getRowHoToLor() const;
```
srwopschall: Minor: these just get the operator off the context_ member struct below?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+   * @brief Construct a new JacobianContributions object
+   *
+   * @param blocks List of {row_space, col_space} pairs defining the Jacobian blocks
+   */
+  JacobianContributions( std::initializer_list<std::pair<BlockSpace, BlockSpace>> blocks );
+
+  /**
+   * @brief Reserve memory for each block contribution
+   *
+   * @param n_pairs Number of interface pairs
```
srwopschall: These comments are clear as to what the args are. Consider mapping this to the classes above where more clarity is needed.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+   * @return std::unique_ptr<LogicalJacobianBlockOp>
    */
-  std::unique_ptr<mfem::BlockOperator> GetMfemBlockJacobian(
-      const MethodData& method_data, const std::vector<std::pair<int, BlockSpace>>& row_info,
-      const std::vector<std::pair<int, BlockSpace>>& col_info ) const;
+  std::unique_ptr<LogicalJacobianBlockOp> BuildJacobianBlockOp(
+      const std::vector<ComputedElementData>& contributions ) const;
+
+  /**
+   * @brief Returns a Jacobian as a single ParSparseMat
```
srwopschall: Minor: "Assemble and return a Jacobian as a single ParSparseMat" ? It looks like it creates one from the contributions input.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+  /**
+   * @brief Access the cached HO->LOR LM (pressure/gap) transfer matrix (DOF-level)
+   *
+   * @note Returns nullptr if LOR is not active or UpdateJacobianXfer() has not been called.
+   */
+  const shared::ParSparseMat* GetLagrangeMultiplierHoToLorTransfer() const;
 
  private:
+  /**
+   * @brief Map legacy Tribol block spaces to solver-facing logical roles and surface partitions
```
srwopschall: Surface partitions is unclear and overloaded. I can't tell if you mean a redecomp "partition" of the surface or something else. I looked at the routine and it looks like the "partition" is denoted as mortar or nonmortar, so let's work on a more clear name for this like surface type, or surface space, or surface designation etc .

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+   */
+  static ContributionSpace ToContributionSpace( BlockSpace block_space );
+
+  /**
+   * @brief Build the DOF map that injects submesh DOFs into the parent FE space
+   */
+  static shared::ParSparseMat BuildSubmeshParentTransferMatrix(
```
srwopschall: Let's discuss this design. I made a comment about not seeing how the operators are necessarily constructed, e.g. the matrix, in the class itself, and now see member routines on MfemData that appear to build these operators, so this class wraps and provided functionality that the operator class doesn't have. (if I'm seeing this correctly).

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+   * @brief Build a DOF-level transfer matrix from a higher-order FE space to a LOR FE space.
+   *
+   * This constructs a sparse matrix representation matching MFEM's
+   * mfem::L2ProjectionGridTransfer::L2ProjectionH1Space::Mult() behavior, i.e.
+   * y = P_lor * R_true * P_ho^T * x, where P_* are Dof_TrueDof prolongation
+   * matrices and R_true is the L2-projection restriction operator on true dofs.
+   *
+   * @note Only supports continuous (H1) spaces and mfem::Ordering::byNODES.
+   */
+  static shared::ParSparseMat BuildLORTransferMatrix( const mfem::ParFiniteElementSpace& ho_fes,
```
srwopschall: Again, would have assumed that building an operator would have been a call to a build routine on the operator class itself.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+                                                      const mfem::ParFiniteElementSpace& lor_fes,
+                                                      const mfem::ParFiniteElementSpace& ho_scalar_fes,
+                                                      const mfem::ParFiniteElementSpace& lor_scalar_fes );
+
   /**
    * @brief Creates and stores data that changes when the redecomp mesh is
    * updated
    */
   struct UpdateData {
+    struct RedecompRoute {
```
srwopschall: "Route" is confusing to me. Add doxygen as to what this struct is intended to do.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+      /// Logical role of the assembled row space after Tribol block aggregation
```
srwopschall: Clarify meaning of "block aggregation". There is a lot of new language in these classes that appear to imply operations etc. that maybe could be clarified.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+      VariableRole row_role;
+      /// Logical role of the assembled column space after Tribol block aggregation
+      VariableRole col_role;
+      /// Local row size of the intermediate surface matrix produced by this route
```
srwopschall: Clarify what an intermediate surface matrix is. Is this an operator? What makes it intermediate? If it is in fact intermediate, what is the beginning and end destination? And what is a route?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+      int height;
+      /// Local column size of the intermediate surface matrix produced by this route
+      int width;
+      /// Redecomp -> surface transfer for this logical row/column pairing
+      std::unique_ptr<redecomp::MatrixTransfer> transfer;
```
srwopschall: How does this struct relate to a RedecompJacobianStageOp, which also has a MatrixTransfer member pointer on it?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+
     /**
      * @brief Construct a new UpdateData object
      *
      * @param parent_data MFEM data associated with displacement and velocity
      * @param submesh_data MFEM data associated with pressure and gap
      */
-    UpdateData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data );
+    UpdateData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data,
+                const mfem::Array<HYPRE_BigInt>& submesh2parent_vdof_list );
```
srwopschall, line 2024: Add doxygen for new mfem::Array input arg.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
 
     /**
-     * @brief Redecomp to parent-linked boundary submesh transfer operators
-     *
-     * @note Indexed by (row_block, col_block)
+     * @brief Redecomp -> surface stage routes used for Jacobian assembly
```
srwopschall: This needs clarifying? Redecomp is a surface mesh, so what does redecomp->surface mean and what is a stage route.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1665,40 +1956,113 @@ class MfemJacobianData {
...
+     * @brief Optional HO->LOR transfer operators used to map Jacobians back to HO spaces
+     */
+    std::unique_ptr<HoToLorTransferOp> primary_ho_to_lor_;
+    std::unique_ptr<HoToLorTransferOp> dual_ho_to_lor_;
+
+    /**
+     * @brief Submesh -> parent transfer operator for primary variables
      */
-    Array2D<std::unique_ptr<redecomp::MatrixTransfer>> submesh_redecomp_xfer_;
+    std::unique_ptr<SubmeshParentTransferOp> submesh_parent_xfer_;
```
srwopschall: Is this only used for primary variables? Do we want to be explicit in the naming to mark a contrast from any operator that may treat dual variables?

src/tribol/mesh/MfemData.cpp
```diff
@@ -20,6 +20,762 @@
 
 namespace tribol {
 
+namespace {
+
+bool GetEnvBool( const char* name )
```
srwopschall: I didn't see this in a .hpp file. Can you add a comment as to what this is doing or meant to do?

src/tribol/mesh/MfemData.cpp
```diff
@@ -20,6 +20,762 @@
...
+{
+  class MfemTrueRestrictionAccessor : public mfem::L2ProjectionGridTransfer::L2ProjectionH1Space {
+   public:
+    MfemTrueRestrictionAccessor( const mfem::ParFiniteElementSpace& pfes_ho,
+                                 const mfem::ParFiniteElementSpace& pfes_lor, bool use_ea )
+        : mfem::L2ProjectionGridTransfer::L2ProjectionH1Space( pfes_ho, pfes_lor, use_ea )
+    {
+    }
+
+    const mfem::Operator* TrueRestriction() const { return R.get(); }
```
srwopschall, line 39: "TryGet" makes it sound like something should be printed if it couldn't get the restriction operator?

src/tribol/mesh/MfemData.cpp
```diff
@@ -20,6 +20,762 @@
...
+    mfem::FiniteElementSpace::ListToMarker( x_vdofs_list, fes.GetVSize(), x_vdofs_marker );
+    R_mat->BooleanMult( x_vdofs_marker, X_vdofs_marker );
+    mfem::FiniteElementSpace::MarkerToList( X_vdofs_marker, tdofs_list );
+  } else {
+    tdofs_list.SetSize( fes.GetNDofs() );
+    fes.GetVDofs( vdim, tdofs_list );
+  }
+}
+
+shared::ParSparseMat BuildTrueDofExtract( const mfem::ParFiniteElementSpace& vec_fes,
```
srwopschall, line 74: If these are helper functions that aren't in the header add doxygen where appropriate.

src/tribol/mesh/MfemData.cpp
```diff
@@ -20,6 +20,762 @@
...
+  }
+  ho2lor.MakeJ();
+  for ( int ilor = 0; ilor < nel_lor; ++ilor ) {
+    const int iho = cf_tr.embeddings[ilor].parent;
+    ho2lor.AddConnection( iho, ilor );
+  }
+  ho2lor.ShiftUpI();
+}
+
+void ElemMixedMass( mfem::Geometry::Type geom, const mfem::FiniteElement& fe_ho, const mfem::FiniteElement& fe_lor,
```
srwopschall: What is this routine doing? Forming an element mixed mass matrix? does it form an element mixed mass matrix and then operate on one of the input mfem::FiniteElement objects?

srwopschall (commented):
@ebchin - Let's talk about MfemData.hpp from a design and clarity standpoint. There are a number of things that weren't as clear as they maybe could be.

src/tribol/mesh/MfemData.hpp
ebchin: Add examples of what Primary, Dual are.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
   bool use_device_;
 };
 
+/**
+ * @brief Solver-facing role of a variable space
+ */
+enum class VariableRole
```
ebchin: VariableType

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+{
+  Primary,
+  Dual
+};
+
+/**
+ * @brief Surface partition associated with a contribution source
+ */
+enum class SurfacePartition
```
ebchin: SurfaceRegion

src/tribol/mesh/MfemData.hpp
ebchin: Consider removal

src/tribol/mesh/MfemData.hpp
ebchin: Break this down into 2 classes: assembled transfer operators and matrix-free/partial-assembly operators.

src/tribol/mesh/MfemData.hpp
ebchin: Get rid of `Impl` functions

src/tribol/mesh/MfemData.hpp
ebchin: BlockPairData

src/tribol/mesh/MfemData.hpp
ebchin: `RedecompMatrixTransfer`

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,286 @@ class MfemSubmeshData {
...
+  const redecomp::MatrixTransfer& transfer_;
+  std::vector<ComputedElementData> contributions_;
+  const Array1D<int>& mortar_elem_map_;
+  const Array1D<int>& nonmortar_elem_map_;
+};
+
+/**
+ * @brief Solver-visible logical Jacobian block assembled from one stage operator
+ */
+struct LogicalBlockTransferContext {
```
ebchin: JacobianTransferInfo

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
   bool use_device_;
 };
 
+/**
+ * @brief HO -> LOR transfer action (matrix-free apply) for one FE-space pair
+ *
+ * This is intentionally matrix-free: it only provides Mult() and does not
+ * expose any assembled operator.
+ */
+class HoToLorTransferAction : public mfem::Operator {
```
srwopschall, line 1665: This class looks a lot better and the comments are super clear. As a sanity check, we don't need the lor factor stored on this class?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+ *
+ * This type is assembled-only and returns an explicit HypreParMatrix wrapper.
+ * It does not cache assembled results.
+ */
+class HoToLorTransferMat {
+ public:
+  /**
+   * @brief Construct a builder that maps HO DOFs to LOR DOFs for one field space
+   *
+   * @param ho_fes Higher-order (HO) finite element space on the surface
```
srwopschall: Minor: clarify what is meant by surface here. Is this meant to be used to transfer Ho submesh field to Lor submesh field? Does this only work with a submesh?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+  const mfem::ParFiniteElementSpace& lor_fes_;
+  /// MFEM HO->LOR transfer object (constructed lazily but reused across Mult() calls)
+  mutable std::unique_ptr<mfem::L2ProjectionGridTransfer> transfer_;
+};
+
+/**
+ * @brief HO -> LOR transfer matrix builder for one FE-space pair
+ *
+ * This type is assembled-only and returns an explicit HypreParMatrix wrapper.
+ * It does not cache assembled results.
```
srwopschall: Sanity check: before we had like three options to assemble(), to assemble and cache, or to just get or assemble, cache and get. Do we want to cache the matrix on this class, or are most use cases sufficient to just assemble and return the ParSparseMat?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+  /// Low-order refined (LOR) surface finite element space
+  const mfem::ParFiniteElementSpace& lor_fes_;
+  /// Scalar (vdim=1) companion HO space used to build component operators
+  const mfem::ParFiniteElementSpace& ho_scalar_fes_;
+  /// Scalar (vdim=1) companion LOR space used to build component operators
+  const mfem::ParFiniteElementSpace& lor_scalar_fes_;
+};
+
+/**
+ * @brief Submesh -> parent DOF transfer matrix builder
```
srwopschall, line 1730: Add note about whether the submesh is always HO. That is, if we use LOR then is the LOR to HO submesh done elsewhere?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+ */
+class SubmeshParentTransferMat {
+ public:
+  /**
+   * @brief Construct a builder that injects boundary-submesh DOFs into the parent FE space
```
srwopschall, line 1735: Consider clarifying/elaborating on what injecting boundary-submesh DOFs into the parent FE space means practically for this transfer.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+   *
+   * @param submesh_fes Boundary submesh finite element space
+   * @param parent_fes Parent volume finite element space
+   * @param submesh2parent_vdof_list Global parent vdof for each local submesh vdof
```
srwopschall, line 1739: Minor, if there is functionality to help construct this list then consider mentioning it. If not, don't worry about it.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+};
+
+/**
+ * @brief Submesh -> parent DOF transfer matrix builder
+ */
+class SubmeshParentTransferMat {
+ public:
+  /**
+   * @brief Construct a builder that injects boundary-submesh DOFs into the parent FE space
+   *
```
srwopschall, line 1736: For consistency mention that the class does not cache the matrix.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+        col_redecomp_fes( &col_redecomp_fes_in ),
+        row_elem_map( &row_map ),
+        col_elem_map( &col_map )
+  {
+  }
+
+  /**
+   * @brief Reserve packed storage for a batch of element contributions
+   *
+   * @param n_entries Number of element-pair contributions that will be appended.
```
srwopschall: I thought this struct holds Jacobian data for a single element pair. This comment makes it sound like it holds data for multiple element pairs, where n_values_per_entry is the number of scalar Jacobian contributions per a single element pair. Can we clarify and revise? 

Is n_entries meant to be like number of block contributions for single element pair where n_values_per_entry is the number of scalar values for a given entry contribution?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+
+  /**
+   * @brief Append one flattened element Jacobian contribution
+   *
+   * @param row_elem_id Tribol element id for the row side of this contribution.
+   * @param col_elem_id Tribol element id for the column side of this contribution.
+   * @param data Pointer to a contiguous, column-major dense block of Jacobian values.
+   * @param size Number of scalar entries in @p data (typically `n_row_dofs * n_col_dofs`).
+   */
+  void append( int row_elem_id, int col_elem_id, const double* data, int size )
```
srwopschall, line 1829: Could `size` here in principle be different than n_values_per_entry in reserve()? Is there a consistency issue.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+  /**
+   * @brief Number of element contributions stored in this packed block
+   */
+  int numEntries() const { return row_elem_ids.size(); }
+};
+
+/**
+ * @brief Assemble a Jacobian on LOR/submesh surface FE spaces from redecomp element contributions
+ */
+class RedecompJacobianTransfer {
```
srwopschall: Rename for consistency? RedecompJacobianTransferMat? Or more explicit naming following the classes above, though long, would be RedecompToSubmeshJacobianTransferMat. The other classes are explicit in what we are transfering, but maybe we talked about if you are on the Redecomp level then you can only transfer up and it is understood that these operators transfer "up" one level at a time.?

Consider mentioning that the matrix is not cached for consistency.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+ public:
+  /**
+   * @brief Construct one redecomp->surface Jacobian transfer for a logical row/column pairing
```
srwopschall: clarify what "logical" is intended to mean here.

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+   *
+   * @param transfer Redecomp transfer route for the given row/col surface FE spaces
+   * @param contributions Packed element Jacobian contributions for one row/col FE-space pairing
```
srwopschall, line 1858: Minor: Vector of packed element Jacobian contributions for one row/col FE-space contact pair

src/tribol/mesh/MfemData.hpp
```diff
@@ -1645,6 +1656,247 @@ class MfemSubmeshData {
...
+  /// Redecomp transfer route used to map redecomp element contributions onto the surface DOF layout
+  const redecomp::MatrixTransfer& transfer_;
+  /// Packed element Jacobian contributions (possibly aggregated from multiple sources)
+  std::vector<PackedPairJacobianContribs> contributions_;
+};
+
+/**
+ * @brief Assemble a solver-visible Jacobian by composing a LOR/submesh Jacobian with explicit transfer operators
+ *
+ * The transfer operators must be explicit assembled sparse matrices. Each transfer list is an ordered chain that maps
```
srwopschall: "explicitly assembled" ?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1656,49 +1908,102 @@ class MfemJacobianData {
...
   void UpdateJacobianXfer();
 
   /**
-   * @brief Returns a Jacobian as an mfem::BlockOperator
+   * @brief Assemble a Jacobian on the LOR or submesh DOF spaces using redecomp::MatrixTransfer
+   *
+   * This returns the intermediate Jacobian prior to any solver-block mapping. Method-specific code is expected to
+   * compose this matrix with true-dof and optional HO/LOR/submesh-parent transfer operators.
+   *
+   * @note The contributions must correspond to a single logical row/column pairing (e.g. primary-primary).
```
srwopschall: Since you took out that enum, does primary-primary mean anything? Is this defined anywhere? Consider clarifying?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1656,49 +1908,102 @@ class MfemJacobianData {
...
    * @brief Builds new transfer data after a new redecomp mesh has been built
    */
   void UpdateJacobianXfer();
 
   /**
-   * @brief Returns a Jacobian as an mfem::BlockOperator
+   * @brief Assemble a Jacobian on the LOR or submesh DOF spaces using redecomp::MatrixTransfer
+   *
+   * This returns the intermediate Jacobian prior to any solver-block mapping. Method-specific code is expected to
+   * compose this matrix with true-dof and optional HO/LOR/submesh-parent transfer operators.
```
srwopschall: Minor: Is this ordering right or is it HO/submesh/LOR-parent"

src/tribol/mesh/MfemData.hpp
```diff
@@ -1656,49 +1908,102 @@ class MfemJacobianData {
...
-   * block 0), this vector defines the mapping from each Tribol space to its corresponding block row index.
-   * @param col_info List of {block_col_index, BlockSpace} pairs. Similar to row_info, this defines the mapping from
-   * each Tribol space to its corresponding block column index.
-   * @return std::unique_ptr<mfem::BlockOperator>
+   * @note Returns nullptr if LOR is not active or UpdateJacobianXfer() has not been called.
    */
-  std::unique_ptr<mfem::BlockOperator> GetMfemBlockJacobian(
-      const MethodData& method_data, const std::vector<std::pair<int, BlockSpace>>& row_info,
-      const std::vector<std::pair<int, BlockSpace>>& col_info ) const;
+  const HoToLorTransferMat* GetDisplacementHoToLorTransferMat() const;
```
srwopschall: Just to clarify, we removed the enum marking a primary or a dual field and all that stuff, so does a HoToLorTransferMat "know" that it is for a displacement field vs. a lagrange multiplier pressure field simply because of the finite element spaces and transfer operator defined?

src/tribol/mesh/MfemData.hpp
```diff
@@ -1656,49 +1908,102 @@ class MfemJacobianData {
...
+    UpdateData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data,
+                const mfem::Array<HYPRE_BigInt>& submesh2parent_vdof_list );
 
     /**
-     * @brief Redecomp to parent-linked boundary submesh transfer operators
-     *
-     * @note Indexed by (row_block, col_block)
+     * @brief Optional HO->LOR transfer operators used to map Jacobians back to HO spaces
+     */
+    std::unique_ptr<HoToLorTransferMat> primary_ho_to_lor_;
```
srwopschall: Are these used to store the displacement transfer matrix and the lagrange multiplier transfer matrix that are returned with the "getters" above?

src/tribol/mesh/MfemData.cpp
```diff
@@ -20,6 +20,302 @@
 
 namespace tribol {
 
+namespace {
+
+std::unique_ptr<shared::ParSparseMat> TryGetMfemTrueRestrictionMatrix(
```
srwopschall, line 25: Are the following routines helper routines? Maybe just add a top comment "Helper routines" to orient

src/tribol/mesh/MfemData.cpp
```diff
@@ -839,234 +1145,201 @@ MfemJacobianData::MfemJacobianData( const MfemMeshData& parent_data, const MfemS
...
 void MfemJacobianData::UpdateJacobianXfer()
 {
-  update_data_ = std::make_unique<UpdateData>( parent_data_, submesh_data_ );
+  update_data_ = std::make_unique<UpdateData>( parent_data_, submesh_data_, submesh2parent_vdof_list_ );
 }
 
-std::unique_ptr<mfem::BlockOperator> MfemJacobianData::GetMfemBlockJacobian(
-    const MethodData& method_data, const std::vector<std::pair<int, BlockSpace>>& row_info,
-    const std::vector<std::pair<int, BlockSpace>>& col_info ) const
+MfemJacobianData::UpdateData::UpdateData( const MfemMeshData& parent_data, const MfemSubmeshData& submesh_data,
```
srwopschall, line 1229: Clarify the two UpdateData....I missed where this was.

src/tribol/mesh/MfemData.cpp
```diff
@@ -839,234 +1145,201 @@ MfemJacobianData::MfemJacobianData( const MfemMeshData& parent_data, const MfemS
...
+                                 col_surface_fes->GlobalVSize(), row_surface_fes->GetDofOffsets(),
+                                 col_surface_fes->GetDofOffsets(), std::move( empty_diag ) );
+  }
 
-      // Pick transfer object
-      redecomp::MatrixTransfer* xfer = GetUpdateData().submesh_redecomp_xfer_( r_blk, c_blk ).get();
+  // Build the redecomp transfer route for this call
+  redecomp::MatrixTransfer transfer( *row_surface_fes, *col_surface_fes, *row_redecomp_fes, *col_redecomp_fes );
+  RedecompJacobianTransfer redecomp_xfer( transfer, contributions );
+  return redecomp_xfer.Assemble();
```
srwopschall: Minor: but i got confused about what is an assembled transfer operator vs. what is the honest to goodness Jacobian matrix. It looks like you create a transfer operator class RedecompJacobianTransfer that takes a redecomp::MatrixTransfer object and contributions and then assemble the Jacobian. Consider for clarity renaming `Assemble()` to `AssembleJacobian()` to be very clear that this is not assembling some transfer matrix that is meant to be chained together in the route back up to the parent space, but is actually assembling the honest to goodness Jacobian at that level. Is this correct? 

If I have that all wrong then rename this routine to AssembleLorOrSubmeshJacobianTransferMat(), because it would distinguish between assembling the Jacobian itself vs assembling a transfer matrix that can get you the Jacobian.

srwopschall (commented):
@ebchin - This is a lot cleaner and clearer. I have some clarifying comments for MfemData.hpp/cpp in this review. Moving on to another chunk.

src/redecomp/transfer/MatrixTransfer.hpp
```diff
@@ -86,6 +86,26 @@ class MatrixTransfer {
                                                const axom::Array<int>& trial_elem_idx,
                                                const axom::Array<mfem::DenseMatrix>& src_elem_mat ) const;
 
+  /**
+   * @brief Transfers element RedecompMesh matrices to parent mfem::ParMesh
+   *
+   * @param test_elem_idx List of element IDs on the redecomp test space
+   * @param trial_elem_idx List of element IDs on the redecomp trial space
+   * @param src_elem_mat_data Flattened array of element-level dense matrices from the redecomp mesh
```
srwopschall: Is this an array of element dense matrices or is it element dense matrices flattened to a 1D array? If the latter consider rewording "Flattened array of element-level dense matrix data from the redecomp mesh" (or something like that).

src/redecomp/transfer/MatrixTransfer.hpp
```diff
@@ -101,6 +121,60 @@ class MatrixTransfer {
   shared::ParSparseMat ConvertToParSparseMat( mfem::SparseMatrix&& sparse, bool parallel_assemble = true ) const;
 
  private:
+  struct CommunicationData {
```
srwopschall, line 134: Minor: doxygen on what this is (although it is clear from the name).

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -11,9 +11,69 @@
...
 namespace redecomp {
 
+void MatrixTransfer::validateTransferInputs( const axom::Array<int>& test_elem_idx,
+                                             const axom::Array<int>& trial_elem_idx, int data_size,
+                                             const char* data_name ) const
+{
+  SLIC_ERROR_IF( test_elem_idx.size() != trial_elem_idx.size() || test_elem_idx.size() != data_size,
+                 axom::fmt::format( "Element index arrays and {} must be the same size.", data_name ) );
+  for ( int i{ 0 }; i < test_elem_idx.size(); ++i ) {
+    SLIC_ERROR_IF( test_elem_idx[i] < 0, "Invalid primary index value." );
```
srwopschall: For all these error messages can you elaborate on where in the code this is failing? If I just see "Invalid primary index value" I'd be so confused as to where this is occuring and what this means.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -11,9 +11,69 @@
...
+void MatrixTransfer::validateDenseMatrices( const axom::Array<int>& test_elem_idx,
+                                            const axom::Array<int>& trial_elem_idx,
+                                            const axom::Array<mfem::DenseMatrix>& src_elem_mat ) const
+{
+  validateTransferInputs( test_elem_idx, trial_elem_idx, src_elem_mat.size(), "element Jacobian contribution array" );
+  for ( int i{ 0 }; i < src_elem_mat.size(); ++i ) {
+    auto n_test_elem_vdofs = redecomp_test_fes_.GetFE( test_elem_idx[i] )->GetDof() * redecomp_test_fes_.GetVDim();
+    auto n_trial_elem_vdofs = redecomp_trial_fes_.GetFE( trial_elem_idx[i] )->GetDof() * redecomp_trial_fes_.GetVDim();
+
+    SLIC_ERROR_IF( src_elem_mat[i].Height() != n_test_elem_vdofs,
```
srwopschall: Again, elaborate the message to indicate where/when/how this is failing. It looks like these messages just indicate how with no other context.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -11,9 +11,69 @@
...
+                                           const axom::Array<int>& src_elem_mat_offsets ) const
+{
+  validateTransferInputs( test_elem_idx, trial_elem_idx, src_elem_mat_offsets.size(),
+                          "element Jacobian offsets array" );
+  for ( int i{ 0 }; i < test_elem_idx.size(); ++i ) {
+    auto n_test_elem_vdofs = redecomp_test_fes_.GetFE( test_elem_idx[i] )->GetDof() * redecomp_test_fes_.GetVDim();
+    auto n_trial_elem_vdofs = redecomp_trial_fes_.GetFE( trial_elem_idx[i] )->GetDof() * redecomp_trial_fes_.GetVDim();
+    auto expected_size = n_test_elem_vdofs * n_trial_elem_vdofs;
+
+    SLIC_ERROR_IF( src_elem_mat_offsets[i] < 0 || src_elem_mat_offsets[i] + expected_size > src_elem_mat_data.size(),
```
srwopschall: Same...add more context that would help a user debug their issue.

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -136,14 +160,168 @@ mfem::SparseMatrix MatrixTransfer::TransferToParallelSparse( const axom::Array<i
...
+                                                         bool parallel_assemble ) const
+{
+  TRIBOL_MARK_FUNCTION;
+  validateFlatMatrices( test_elem_idx, trial_elem_idx, src_elem_mat_data, src_elem_mat_offsets );
+
+  auto test_redecomp = dynamic_cast<const RedecompMesh*>( redecomp_test_fes_.GetMesh() );
+  auto trial_redecomp = dynamic_cast<const RedecompMesh*>( redecomp_trial_fes_.GetMesh() );
+  auto comm_data = buildCommunicationData( *test_redecomp, *trial_redecomp, test_elem_idx, trial_elem_idx );
+
+  TRIBOL_MARK_BEGIN( "SetupTriplets" );
```
srwopschall, line 215: Consider elaborating what a triplet is in this case? SetupRowColValTriplets?

src/redecomp/transfer/MatrixTransfer.cpp
```diff
@@ -136,14 +160,168 @@ mfem::SparseMatrix MatrixTransfer::TransferToParallelSparse( const axom::Array<i
...
+  // Sort triplets by row then column
+  std::sort( triplets.begin(), triplets.end(), []( const Triplet& a, const Triplet& b ) {
+    if ( a.row != b.row ) return a.row < b.row;
+    return a.col < b.col;
+  } );
+  TRIBOL_MARK_END( "SortTriplets" );
+
+  TRIBOL_MARK_BEGIN( "CSRConstruction" );
+  // Count non-zeros and merge duplicates
+  int num_unique_nonzeros = 0;
```
srwopschall, line 294: doesn't look like the sort and merge code here can be tested in a unit test, but just make sure this stuff is solid.

src/shared/math/ParSparseMat.hpp
```diff
@@ -272,6 +272,21 @@ class ParSparseMat : public ParSparseMatView {
...
 
+  /**
+   * @brief Construct from MPI communicator, global size, row/column starts, and mfem::SparseMatrix rvalue
+   *
+   * @param comm MPI communicator
+   * @param global_num_rows Global number of rows
+   * @param global_num_cols Global number of columns
+   * @param row_starts Global row partitioning
+   * @param col_starts Global column partitioning
+   * @param diag Local diagonal block SparseMatrix (rvalue)
```
srwopschall, line 295: Can you clarify diag? In other constructors we can construct a diagonal matrix where diag is the value for diagonal entries. Here is just looks like we are constructing a ParSpareMat from data held in a mfem::SparseMatrix? If that is the case does this need to be so specific as to stating that that mfem sparse matrix is specifically a diagonal block? Can't it really be any block?

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -0,0 +1,1021 @@
...
+class MfemJacobianTest : public testing::Test {
+ protected:
+  void SetUp() override {}
+
+  void TearDown() override { tribol::finalize(); }
+};
+
+namespace {
+
+enum class FieldKind
```
srwopschall, line 43: These enums look like the ones we got rid of in the new MfemData.hpp code for the transfer operators. Do we still need these here or can this be simplified with the refactor since last review?

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -0,0 +1,1021 @@
...
+  diff += y_mfem;
+  diff -= y_tribol;
+
+  double local_norm2 = diff * diff;
+  double global_norm2 = 0.0;
+  MPI_Allreduce( &local_norm2, &global_norm2, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD );
+  EXPECT_LT( std::sqrt( global_norm2 ), 1e-10 );
+}
+
+double MaxAbsMatrixEntry( const shared::ParSparseMatView& mat )
```
srwopschall, line 182: Any value in pulling some of these utility functions out of the test and placing in the main src?

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -0,0 +1,1021 @@
...
+        MPI_Comm_size( MPI_COMM_WORLD, &n_ranks );
+        if ( n_ranks == 1 ) {
+          FAIL() << "No surface elements found on mesh 1 in serial run.";
+        }
+      }
+    } );
+  }
+}
+
+TEST_F( MfemJacobianTest, logical_jacobian_primary_dual_assembles_in_ho_and_lor )
```
srwopschall: Looks like codex language with "logical". Consider clarity.

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -0,0 +1,1021 @@
...
+        EXPECT_GT( combined->NNZ(), 0 );
+      }
+
+      auto diff = shared::ParSparseMatView( &combined.get() ) - shared::ParSparseMatView( &expected.get() );
+      EXPECT_LT( MaxAbsMatrixEntry( shared::ParSparseMatView( &diff.get() ) ), 1e-12 );
+    } );
+  }
+}
+
+TEST_F( MfemJacobianTest, mfem_block_jacobian_preserves_block_values_and_layout )
```
srwopschall, line 806: Minor, looks like there are tests that test assembling a single block and there is a helper routine in this file along those lines. Is there a test that assembles the full Jacobian? Note: this was a lot of code so I breezed through pretty quickly just to get the idea of what was being tested.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -7,11 +7,233 @@
...
+  mfem::Array<int> offsets( max_block + 2 );
+  offsets[0] = 0;
+  offsets[1] = primary_size;
+  if ( max_block > 0 ) {
+    offsets[2] = primary_size + dual_size;
+  }
+  return offsets;
+}
+
+mfem::Array<int> ComputeSingleMortarInactiveDualTrueDofs( const MfemMeshData& mesh_data,
```
srwopschall, line 36: Single mortar is a method and this looks like a new API function? Or a helper function? I'm hesitant to commit method specific naming in mfem_tribol.cpp

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -7,11 +7,233 @@
...
+  mfem::Array<int> mortar_tdof_marker( submesh_fe_space.GetTrueVSize() );
+  submesh_fe_space.GetRestrictionMatrix()->BooleanMult( mortar_dof_marker, mortar_tdof_marker );
+
+  // Convert markers of tdofs only on mortar surface to a list.
+  mfem::Array<int> mortar_tdof_list;
+  mfem::FiniteElementSpace::MarkerToList( mortar_tdof_marker, mortar_tdof_list );
+  return mortar_tdof_list;
+}
+
+std::vector<PackedPairJacobianContribs> BuildPackedPairJacobianContribs(
```
srwopschall, line 78: Minor: Should a routine like this (and/or others) be inside the API file?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -7,11 +7,233 @@
...
+      data.append( row_elem_ids[i], col_elem_ids[i], J_block[i].GetData(), size );
+    }
+
+    computed.push_back( std::move( data ) );
+  }
+
+  return computed;
+}
+
+std::unique_ptr<mfem::BlockOperator> BuildMfemBlockJacobian( const CouplingScheme& cs, const MethodData& method_data,
```
srwopschall, line 163: Does this need to be exposed to the host code? Should this routine be in mfem data?

srwopschall (commented):
@ebchin - final block of review comments. Take a look and let's discuss.

src/redecomp/transfer/MatrixTransfer.cpp
ebchin: Updated comments in this file

src/shared/math/ParSparseMat.hpp
ebchin: This is related to how hypre packages sparse matrices.  Longer explanatory comment added.

src/tests/tribol_mfem_jacobian.cpp
ebchin: They're used pretty extensively to signal what is being tested in each test. We can remove it, but I'd say this is better to keep.

src/tests/tribol_mfem_jacobian.cpp
ebchin: Maybe if we need it somewhere else in the future

src/tests/tribol_mfem_jacobian.cpp
ebchin: Ahh these snuck through :)

src/tests/tribol_mfem_jacobian.cpp
ebchin: These first tests are more unit-test-y. The last one is probably what you're looking for.

src/tribol/interface/mfem_tribol.cpp
ebchin: It's compatibility code so the old mortar path still works. What we were calling GetMfemBlockJacobian() before was probably more accurately GetMfemSingleMortarJacobian(), so I'm just being more explicit this time around.

src/tribol/mesh/MfemData.cpp
ebchin: Yeah, they're helpers.  I think the anonymous namespace is enough to indicate this?

src/tribol/mesh/MfemData.cpp
ebchin: It's the constructor for the MfemJacobianData::UpdateData class

src/tribol/mesh/MfemData.cpp
ebchin: I like the naming consistent (i.e. calling Assemble() will give you a shared::ParSparseMat of whatever the operator does). I'll change the name of the class to RedecompJacobianAssembler

src/tribol/mesh/MfemData.hpp
ebchin: No, the finite element space is needed.

src/tribol/mesh/MfemData.hpp
ebchin: I like to let the caller decide what to do with it, rather than caching it without anyone asking it to do that.

src/tribol/mesh/MfemData.hpp
ebchin: It's only used in Tribol for a surface, but the method is general.

src/tribol/mesh/MfemData.hpp
ebchin: The submesh could be HO or linear. Let's keep this the way it is since it's consistent with the naming for the meshes elsewhere in Tribol (parent mesh, submesh, LOR mesh, redecomp mesh, Tribol mesh).

src/tribol/mesh/MfemData.hpp
ebchin: Yeah, "inject" is a codex choice, but I think it's relatively descriptive for how the operator works (it is an injective linear map, see e.g. https://en.wikipedia.org/wiki/Injective_function)

src/tribol/mesh/MfemData.hpp
ebchin: See line 1144 for where it's constructed

src/tribol/mesh/MfemData.hpp
ebchin: No it holds all the pair contributions. The struct name has been changed to try to indicate this. Let me know what we might need to change it to to make it clearer.  n_entries is number of pairs that will be added to the struct, n_values_per_entry is the number of Jacobian entries per pair.

src/tribol/mesh/MfemData.hpp
ebchin: This is for all the pair contributions, not one. So we are reserve()ing the total number of Jacobian entries we want.

src/tribol/mesh/MfemData.hpp
ebchin: Renamed RedecompJacobianAssembler

src/tribol/mesh/MfemData.hpp
ebchin: It's for all the pairs for a given row/col

src/tribol/mesh/MfemData.hpp
ebchin: removed

src/tribol/mesh/MfemData.hpp
ebchin: Pairing is encoded by PackedPairJacobianContribs now

src/tribol/mesh/MfemData.hpp
ebchin: Yeah, the stuff in PackedPairJacobianContribs is enough.

src/tribol/mesh/MfemData.hpp
ebchin: No, the assembled matrices are not stored. Just the "recipe" here.

src/tribol/mesh/MfemData.hpp
srwopschall: Fixed to be more specific of what surface representation we have.

src/tribol/mesh/MfemData.cpp
srwopschall: Let's tease apart a redecomp-to-lor transfer operator class OR jacobian assembler class and another one for redecomp-to-submesh and which one is used is deterimed "higher-up".

srwopschall (approved):
@ebchin - Looks good. Just separate the redecomp to lor or to submesh classes.

PR #197, opened 2026-02-23
Enable enforcement of non-zero gaps

ebchin:
Adds a residual gap parameter that allows enforcement of non-zero gaps. Adds changes to binning and redecomp ghost layer to support non-zero gaps.


src/tribol/common/Parameters.hpp
```diff
@@ -507,6 +507,7 @@ struct Parameters {
 
   EnforcementLocation enforcement_location =
       EnforcementLocation::QuadraturePoint;  ///! Defaults to quadrature-point enforcement
+  RealT residual_gap = 0.0;                  ///! User defined residual gap constraint
```
srwopschall: Let's add a note that positive is out of contact per Tribol definition (please verify this) and that negative is what is traditionally in contact or interpenetrated. Do we want to guard against negative values of residual_gap or give the user ultimate control over this as it sounds like a niche application?

src/tribol/geom/CompGeom.hpp
```diff
@@ -1034,7 +1034,7 @@ TRIBOL_HOST_DEVICE inline bool CommonPlanePair::exceedsMaxAutoInterpen( const Me
     RealT max_interpen = -1. * params.auto_contact_pen_frac *
                          axom::utilities::min( mesh1.getElementData().m_thickness[faceId1],
                                                mesh2.getElementData().m_thickness[faceId2] );
-    if ( gap < max_interpen ) {
+    if ( ( gap - params.residual_gap ) < max_interpen ) {
```
srwopschall, line 1037: I think this is ok, but just confirming that with residual_gap = 0.0, this checks to see if the gap is more than 30% the smallest element thickness. Here, if residual_gap is let's say 1.0, this checks to see if you have blown past the contact gap constraint by more than 30% the smallest element thickness. For residual_gap >= 0.0 this is the intent. If, however, residual_gap < 0.0, then even interpenetration more than 30% of the smallest element thickness could get you into trouble for single element thick thin structures, which this check is meant to guard against contact happening between faces on opposite sides of thin structures. For that reason I would consider only residual_gap >= 0.0 for auto-contact applications.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -775,9 +775,11 @@ void updateMfemParallelDecomposition( int n_ranks, bool force_new_redecomp )
       if ( mfem_data->GetLORFactor() > 1 ) {
         effective_binning_proximity *= static_cast<RealT>( mfem_data->GetLORFactor() );
       }
+      auto residual_gap = cs.getParameters().residual_gap;
       // creates a new redecomp mesh based on updated coordinates (if criteria is met) and updates transfer operators
       // and displacement, velocity, and response grid functions based on new redecomp mesh
-      auto new_redecomp = mfem_data->UpdateMfemMeshData( effective_binning_proximity, n_ranks, force_new_redecomp );
+      auto new_redecomp =
+          mfem_data->UpdateMfemMeshData( effective_binning_proximity, n_ranks, force_new_redecomp, residual_gap );
```
srwopschall, line 782: Is this how a host-code "registers" the residual_gap? Or do we have an API function to register it once? Any reason you'd change it between subsequent calls to UpdateMfemMeshData()?

src/tribol/interface/tribol.cpp
```diff
@@ -352,6 +352,22 @@ void setBinningProximityScale( IndexT cs_id, RealT binning_proximity_scale )
 
 }  // end setBinningProximityScale()
 
+//------------------------------------------------------------------------------
+void setResidualGap( IndexT cs_id, RealT residual_gap )
```
srwopschall, line 356: Ah, here it is. Do we have to have residual_gap as an argument to UpdateMfemMeshData() if a host-code has to call this? Or does UpdateMfemMeshData() call this routine to set it? Let's figure out the best path.

src/tribol/interface/tribol.cpp
```diff
@@ -352,6 +352,22 @@ void setBinningProximityScale( IndexT cs_id, RealT binning_proximity_scale )
...
+{
+  auto cs = CouplingSchemeManager::getInstance().findData( cs_id );
+
+  // check to see if coupling scheme exists
+  SLIC_ERROR_ROOT_IF(
+      !cs, "tribol::setResidualGap(): call tribol::registerCouplingScheme() " << "prior to calling this routine." );
+
+  cs->getParameters().residual_gap = residual_gap;
```
srwopschall, line 365: Should we somehow check to see if residual_gap has been previously registered and it is now being changed and issue a warning? Maybe that is overkill.

src/tribol/mesh/CouplingScheme.hpp
```diff
@@ -1050,8 +1050,9 @@ TRIBOL_HOST_DEVICE inline RealT CouplingScheme::Viewer::getGapTol( IndexT fid1,
           break;
 
         default:
-          gap_tol = -1. * m_parameters.gap_tol_ratio *
-                    axom::utilities::max( m_mesh1.getFaceRadius()[fid1], m_mesh2.getFaceRadius()[fid2] );
+          gap_tol = m_parameters.residual_gap -
```
srwopschall, line 1056: Consider a comment for clarity where negative gap_tol is for classic interpenetration and a positive gap_tol is for separation kinematic constraints (e.g. residual_gap > 0.0)

src/tribol/mesh/MfemData.cpp
```diff
@@ -721,7 +721,8 @@ void MfemMeshData::SetParentReferenceCoords( const mfem::ParGridFunction& refere
   }
 }
 
-bool MfemMeshData::UpdateMfemMeshData( RealT binning_proximity_scale, int n_ranks, bool force_new_redecomp )
+bool MfemMeshData::UpdateMfemMeshData( RealT binning_proximity_scale, int n_ranks, bool force_new_redecomp,
+                                       RealT residual_gap )
```
srwopschall, line 725: Same...clarify if residual_gap is best passed as an argument in this function.

src/tribol/mesh/MfemData.cpp
```diff
@@ -1046,19 +1047,19 @@ MfemMeshData::UpdateData::UpdateData( mfem::ParSubMesh& submesh, mfem::ParMesh*
...
                                       RealT binning_proximity_scale, int n_ranks, int allocator_id,
-                                      RealT redecomp_trigger_displacement )
+                                      RealT redecomp_trigger_displacement, RealT residual_gap )
     : redecomp_mesh_{ lor_mesh
                           ? redecomp::RedecompMesh(
                                 *lor_mesh,
                                 binning_proximity_scale * redecomp::RedecompMesh::MaxElementSize(
                                                               *lor_mesh, redecomp::MPIUtility( lor_mesh->GetComm() ) ) +
-                                    redecomp_trigger_displacement,
+                                    redecomp_trigger_displacement + residual_gap,
```
srwopschall, line 1056: question...the redecomp_trigger_displacement is more for transverse displacement right? This is to catch if faces have slid past the "fuzz" or ghost faces on the other surface? Residual gap is a "normal displacement". If so, is this correct?

src/tribol/mesh/MfemData.cpp
```diff
@@ -1046,19 +1047,19 @@ MfemMeshData::UpdateData::UpdateData( mfem::ParSubMesh& submesh, mfem::ParMesh*
...
                                 redecomp::RedecompMesh::RCB, n_ranks )
                           : redecomp::RedecompMesh(
                                 submesh,
                                 binning_proximity_scale * redecomp::RedecompMesh::MaxElementSize(
                                                               submesh, redecomp::MPIUtility( submesh.GetComm() ) ) +
-                                    redecomp_trigger_displacement,
+                                    redecomp_trigger_displacement + residual_gap,
```
srwopschall, line 1062: Same question re redecomp_trigger_displacement being a transverse sliding displacement vs. residual gap being a normal displacement. If this is correct, what is it intended to do?

srwopschall: > Same question re redecomp_trigger_displacement being a transverse sliding displacement vs. residual gap being a normal displacement. If this is correct, what is it intended to do?

It definitely seems like you'd want the residual_gap to be incorporated into the redecomp geometric decomposition because you may need to look for "proximate" faces that are actually further apart. If residual_gap is 1.0, you may want to look for faces that are alpha*residual_gap where alpha>1.0 is for faces that haven't violated the contact constraint, but are actually further apart than you'd normally consider if residual_gap = 0.0.

src/tribol/mesh/MfemData.hpp
```diff
@@ -654,11 +654,13 @@ class MfemMeshData {
    * @param n_ranks Number of ranks in the parallel decomposition
    * @param force_new_redecomp If true, construct a new RedecompMesh even if displacement threshold is not met (default
    * = false)
+   * @param residual_gap Residual gap to add to ghost layer thickness (default = 0.0)
```
srwopschall, line 657: I though the ghost layer was on the surface to capture any sliding. Residual gap is a normal gap component so these seem at odds.

src/tribol/physics/ContactFormulation.hpp
```diff
@@ -119,6 +119,11 @@ class ContactFormulation {
    */
   virtual void updateConstantPenaltyStiffness( double /*mesh1_penalty*/, double /*mesh2_penalty*/ ) {}
 
+  /**
+   * @brief Update the residual-gap offset on formulations that cache contact parameters
+   */
+  virtual void updateResidualGap( RealT /*residual_gap*/ ) {}
```
srwopschall: Just wondering if we should preclude or warn against updating this to a new value during the life of a contact formulation? Perhaps this should be "setResidualGap()"? Is there a need to update it for a fixed coupling scheme/formulation?

src/tribol/physics/EnergyMortar.cpp
```diff
@@ -19,11 +19,17 @@ namespace {
...
-  double k{ 1.0 };    // Penalty stiffness
+  int N{ 3 };                  // No. of quadrature points
+  double del{ 0.1 };           // Smoothing parameter
+  double k{ 1.0 };             // Penalty stiffness
+  double residual_gap{ 0.0 };  // User-defined gap offset
 };
 
+TRIBOL_ENZYME_INLINE double effective_gap( double normal_gap, double normal_alignment, double residual_gap )
+{
+  return normal_gap * normal_alignment - residual_gap;
```
srwopschall: What is normal_alignment? Wouldn't normal_gap already be aligned in the "normal" direction?

src/tribol/physics/EnergyMortarAdapter.cpp
```diff
@@ -23,6 +24,7 @@ EnergyMortarAdapter<EnforcementLocation>::EnergyMortarAdapter( MfemMeshData& mes
   params_.del = delta;
   params_.N = N;
   params_.enzyme_quadrature = enzyme_quadrature;
+  params_.residual_gap = residual_gap;
```
srwopschall, line 27: Again, should we as a sanity check preclude negative residual_gaps?

src/tribol/physics/EnergyMortarAdapter.cpp
```diff
@@ -57,6 +59,13 @@ void EnergyMortarAdapter<EnforcementLocation>::updateConstantPenaltyStiffness( d
   evaluator_ = std::make_unique<EnergyMortarCalculator>( params_ );
 }
 
+template <template <typename> class EnforcementLocation>
+void EnergyMortarAdapter<EnforcementLocation>::updateResidualGap( RealT residual_gap )
```
srwopschall: Is this more of a setter function? Or do you need to update it between cycles or tribol update calls?

src/tribol/physics/EnergyMortarAdapter.hpp
```diff
@@ -138,13 +138,14 @@ class EnergyMortarAdapter : public EnforcementLocation<EnergyMortarAdapter<Enfor
    * @param enzyme_quadrature If true, use Enzyme-assisted quadrature
    * @param use_penalty If true, interpret the dual field as pressure; otherwise interpret it as a Lagrange multiplier
    * vector (LM mode)
+   * @param residual_gap User-defined gap offset subtracted from the kinematic gap
```
srwopschall: Similar to somewhere else consider adding that positive enforces a separation gap constraint etc.

src/tribol/physics/EnergyMortarAdapter.hpp
```diff
@@ -210,6 +211,13 @@ class EnergyMortarAdapter : public EnforcementLocation<EnergyMortarAdapter<Enfor
    */
   void updateConstantPenaltyStiffness( double mesh1_penalty, double mesh2_penalty ) override;
 
+  /**
+   * @brief Update residual-gap offset
+   *
+   * @param residual_gap User-defined gap offset
+   */
+  void updateResidualGap( RealT residual_gap ) override;
```
srwopschall: setResidualGap()?

src/tribol/physics/Mortar.cpp
```diff
@@ -163,6 +163,9 @@ void ComputeNodalGap<SINGLE_MORTAR>( SurfaceContactElem& elem )
 
       g1 += dotProd( nrml_a.data(), &elem.faceCoords1[elem.dim * b], elem.dim ) * nab_1;
       g2 += dotProd( nrml_a.data(), &elem.faceCoords2[elem.dim * b], elem.dim ) * nab_2;
+
+      // also integrate the residual gap contribution
+      g2 += residual_gap * nab_2;
```
srwopschall, line 168: I haven't written out the math, but aligned mortar and common plane subtract the residual gap.

src/tribol/physics/Mortar.cpp
```diff
@@ -976,6 +982,9 @@ void ComputeMortarForceEnzyme( const RealT* x1, const RealT* n1, const RealT* p1
     for ( int d{ 0 }; d < 3; ++d ) {
       g1[i] += n1[d * size1 + i] * gap_v[d];
     }
+    for ( int j{ 0 }; j < size1; ++j ) {
+      g1[i] -= residual_gap * mortar_mat1[i * size1 + j];
```
srwopschall, line 986: In contrast to ComputeNodalGap<SINGLE_MORTAR>() here we subtract off the residual_gap component. Please verify/clarify.

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -296,9 +296,10 @@ class GridSearch : public SearchBase {
     // Find matches in first mesh (with index 'fromIdx')
     // with candidate elements in second mesh (with index 'toIdx')
     // int k = 0;  // Debug only
+    RealT residual_gap = m_coupling_scheme->getParameters().residual_gap;
     for ( int toIdx = 0; toIdx < m_mesh2.numberOfElements(); ++toIdx ) {
       SpatialBoundingBox bbox = elementBoundingBox( m_mesh2, toIdx );
-      inflateBBox( bbox, e_binning_proximity_scale );
+      inflateBBox( bbox, e_binning_proximity_scale, residual_gap );
```
srwopschall, line 302: Can you orient me to a test where a residual_gap > 0.0 is tested with the binning?

src/tests/tribol_enzyme_element_mortar.cpp
```diff
@@ -974,6 +974,45 @@ TEST_F( EnzymeElementMortarTest, NoOverlap )
   FDCheck( x1, x2, n1, p1, x1_stencil, x2_stencil, num_nodes, check_scale, len_collapse_ratio );
 }
 
+TEST_F( EnzymeElementMortarTest, ExactOverlapResidualGap )
+{
+  // clang-format off
+  // Setup two overlapping elements with a gap of 0.1
```
srwopschall: Note that 0.1 is separation.

src/tests/tribol_enzyme_element_mortar.cpp
```diff
@@ -974,6 +974,45 @@ TEST_F( EnzymeElementMortarTest, NoOverlap )
...
+  double x2[12] = { 0.0, 0.0, 1.0, 1.0,
+                    0.0, 1.0, 1.0, 0.0,
+                    0.1, 0.1, 0.1, 0.1 };
+  double n1[12] = { 0.0, 0.0, 0.0, 0.0,
+                    0.0, 0.0, 0.0, 0.0,
+                    1.0, 1.0, 1.0, 1.0 };
+  double p1[4] = { 1.0, 1.0, 1.0, 1.0 };
+  // clang-format on
+
+  double residual_gap = 0.15;
```
srwopschall, line 994: So this should trigger contact?

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
                             []( double stiffness ) { return stiffness != 0.0; } ) );
 }
 
+TEST( EnergyMortarResidualGapCheck, AssembledGapIsShiftedByArea )
+{
+  RealT x1[2] = { 0.0, 1.0 };
```
srwopschall, line 440: Add comment about what the actual gap is and whether separation or interpenetration.

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
...
+  MeshData mesh2( 1, 1, 2, conn2, LINEAR_EDGE, x2, y2, nullptr, MemorySpace::Host );
+
+  ContactParams params;
+  params.del = 0.1;
+  params.k = 3.0;
+  params.N = 3;
+  params.enzyme_quadrature = true;
+
+  double gap_without_residual[2] = { 0.0, 0.0 };
+  double area_without_residual[2] = { 0.0, 0.0 };
```
srwopschall: Is this the overlap area without residual gap?

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
...
+
+  params.residual_gap = 0.15;
+  double gap_with_residual[2] = { 0.0, 0.0 };
+  double area_with_residual[2] = { 0.0, 0.0 };
+  EnergyMortarCalculator evaluator_with_residual( params );
+  evaluator_with_residual.compute_gtilde_and_area( InterfacePair( 0, 0 ), mesh1.getView(), mesh2.getView(),
+                                                   gap_with_residual, area_with_residual );
+
+  for ( int i = 0; i < 2; ++i ) {
+    EXPECT_NEAR( area_with_residual[i], area_without_residual[i], 1.0e-14 );
```
srwopschall: We'd expect them close because the projected areas of overlap shouldn't change if you have a nonzero residual gap?

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
...
+    EXPECT_NEAR( gap_with_residual[i], gap_without_residual[i] - params.residual_gap * area_without_residual[i],
```
srwopschall: Clarify what this check is doing by multiplying the residual_gap by the area_without_residual.

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
...
+                 1.0e-14 );
+  }
+}
+
+TEST( EnergyMortarResidualGapCheck, QuadraturePointOpenGapBecomesActive )
+{
+  RealT x1[2] = { 0.0, 1.0 };
```
srwopschall, line 480: Again, add comment as to what the gap is for this configuration.

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -434,6 +434,74 @@ TEST( QuadraturePointPenaltyCheck, ZeroGapRetainsActiveTangent )
...
+  EnergyMortarCalculator evaluator_without_residual( params );
+  const auto inactive = evaluator_without_residual.compute_quadrature_point_penalty_data(
+      InterfacePair( 0, 0 ), mesh1.getView(), mesh2.getView() );
+  EXPECT_EQ( inactive.energy, 0.0 );
+
+  params.residual_gap = 0.15;
+  EnergyMortarCalculator evaluator_with_residual( params );
+  const auto active = evaluator_with_residual.compute_quadrature_point_penalty_data( InterfacePair( 0, 0 ),
+                                                                                     mesh1.getView(), mesh2.getView() );
+  EXPECT_GT( active.energy, 0.0 );
```
srwopschall, line 511: Should the checks be similar to the test above? Can you clarify what the energy check is doing or means for a residual_gap test?

src/tests/tribol_finite_diff_energy_mortar.cpp
```diff
@@ -451,6 +519,7 @@ TEST( QuadraturePointPenaltyCheck, DerivativesMatchFiniteDifference )
   params.k = 3.0;
   params.N = 3;
   params.enzyme_quadrature = true;
+  params.residual_gap = 0.15;
```
srwopschall: Looks like the next few tests just add a nonzero residual_gap. Does this change the nature of the test as compared to the original which implicitly had a residual_gap = 0.0? Does it matter?

srwopschall (commented):
@ebchin - Looks good. Some comments to address and minor things for us to verify/clarify prior to approval.

src/tribol/common/Parameters.hpp
ebchin: verified gap, will reject negative values.

src/tribol/geom/CompGeom.hpp
ebchin: yes, correct

src/tribol/interface/mfem_tribol.cpp
ebchin: This is an internal argument; a host code calls `setResidualGap()`.

src/tribol/interface/tribol.cpp
ebchin: `UpdateMfemMeshData()` is an internal routine. It's called by `updateMfemParallelDecomposition()` which reads the residual gap set by `setResidualGap()`.

src/tribol/interface/tribol.cpp
ebchin: Seems unnecessary. Also, how would you know if it has been previously registered?

src/tribol/mesh/CouplingScheme.hpp
ebchin: added a comment

src/tribol/mesh/MfemData.cpp
ebchin: UpdateMfemMeshData() is an internal routine. It's called by updateMfemParallelDecomposition() which reads the residual gap set by setResidualGap().

src/tribol/mesh/MfemData.cpp
ebchin: Redecomp ghost lengths are in all directions and are aligned with the coordinates (not element directions), so adding it to the fuzz works.

src/tribol/physics/ContactFormulation.hpp
ebchin: renamed

src/tribol/interface/tribol.cpp
srwopschall: There are ways that aren't hard, but I agree it is unnecessary. A simple comment at the API level or note on the intended use case should suffice. Or rather anyone familiar with contact should know not to change the constraint mid-simulation.

src/tribol/mesh/MfemData.cpp
srwopschall: Ah ok this makes sense. I was viewing this relative to one surface, not a fuzz length aligned with the coordinate directions, which makes more sense.

src/tribol/physics/EnergyMortar.cpp
ebchin: Renamed to gap_normal and normal_cosine

src/tribol/physics/EnergyMortarAdapter.cpp
ebchin: Added to public API setter `tribol::setResidualGap()`

src/tribol/physics/EnergyMortarAdapter.cpp
ebchin: Renamed to `setResidualGap()`

src/tribol/physics/EnergyMortarAdapter.hpp
ebchin: Updated comment

src/tribol/physics/Mortar.cpp
ebchin: Added a clarifying comment

src/tribol/physics/Mortar.cpp
ebchin: Verified. It's calculating it in g1 here but in g2 in SINGLE_MORTAR.

src/tribol/search/InterfacePairFinder.cpp
ebchin: Added a new test to tribol_residual_gap.cpp that only passes if the residual_gap is working.

src/tests/tribol_enzyme_element_mortar.cpp
ebchin: Added clarification

src/tests/tribol_enzyme_element_mortar.cpp
ebchin: Yes

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Added comment (separation of 0.1)

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Yes, nodal tributary area contributions for the projected overlap. Renamed to `tributary_area_without_residual` and `tributary_area_with_residual` to be clearer.

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Correct. Added a comment

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Integrated gap is in units gap*area.

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Added

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Added checks to match test above. Energy > 0 means the contact is active

src/tests/tribol_finite_diff_energy_mortar.cpp
ebchin: Good catch -- we want to test non-zero residual gap, but we should keep the original. I implemented the test with both options

srwopschall (approved):

PR #198, opened 2026-02-23
Update MFEM and Axom + build enzyme on toss4 cray systems

chapman39:
https://github.com/llnl/smith/pull/1534

### DIFF
--- a/.gitlab/build_toss4.yml
+++ b/.gitlab/build_toss4.yml
@@ -19,7 +19,7 @@
       when: never
     - when: on_success
   before_script:
-    - module load cmake/3.23.1
+    - module load cmake/3.30.5
 
 
 ####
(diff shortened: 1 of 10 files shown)


chapman39:
tuolumne is currently down which is why its failing CI. this PR is ready for review.

srwopschall (approved):

.gitlab/build_toss4.yml
```diff
@@ -38,15 +38,6 @@ toss4-llvm_19_1_3-src:
     ALLOC_TIME: "30"
   extends: .src_build_on_toss4
 
-toss4-llvm_19_1_3-src-codevelop:
```
srwopschall, line 41: Verify that this can be removed.

chapman39: i removed this pipeline because the codevelop name was misleading. however i notice now that by default tribol builds as Debug, and this test tests tribol Release builds. if we still want to test Release builds, i can re-add this, and give it a more proper name. e.g. `toss4-llvm_19_1_3-src-release`

ebchin: We used to have a no redecomp CI build on toss 4, do you think we could add one back in (in a different PR, potentially) @chapman39?  Preferably with no optional TPLs (i.e. no RAJA, no Umpire, no Enzyme)?

Regarding a release build, since we've seen different Enzyme behavior on release vs. debug builds, I think we also need a release build in CI.  `toss4-llvm_19_1_3-src-release` sounds good to me!

ebchin (approved):

PR #199, opened 2026-03-02
Fix race condition

liu15:
This is technically a race condition if more than one thread set the same value, though in practice it will probably work fine without the atomic.  This is, however, detected by thread sanitizer.


ebchin (approved):
Thanks @liu15

srwopschall (approved):

PR #201, opened 2026-03-28
Workaround for CUDA failure that can occur for optimized code

liu15:
Workaround for a CUDA segfault that only occurs for optimized code.
The longer term solution is probably using an axom reducer.

The issue is referenced in
https://github.com/llnl/axom/issues/1833

### DIFF
--- a/src/tribol/search/InterfacePairFinder.cpp
+++ b/src/tribol/search/InterfacePairFinder.cpp
@@ -127,7 +127,7 @@ class CartesianProduct : public SearchBase {
     contactPairs.resize( countArray_host[0] );
 
     int zero = 0;
-    axom::copy(countArray.data(), &zero, sizeof(int));
+    axom::copy( countArray.data(), &zero, sizeof(int) );
     auto pairs_view = m_coupling_scheme->getInterfacePairs().view();
     // fill proximate pairs array
     forAllExec(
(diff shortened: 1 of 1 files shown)


ebchin:
Thanks @liu15. Can you also add a quick note referencing the axom issue?

liu15:
> Thanks @liu15. Can you also add a quick note referencing the axom issue?

Isn't it in the description?

src/tribol/search/InterfacePairFinder.cpp
```diff
@@ -126,7 +126,8 @@ class CartesianProduct : public SearchBase {
     auto& contactPairs = m_coupling_scheme->getInterfacePairs();
     contactPairs.resize( countArray_host[0] );
 
-    countArray.fill( 0 );
+    int zero = 0;
```
ebchin, line 129:
```suggestion
    // Workaround for axom::Array::fill() issue on rzvector (CUDA). See Axom #1833.
    int zero = 0;
```

Yes, the description is clear, I just meant something like the above, so the rationale is clear in the code and so we can track the issue and it's resolution.

But, as you mention, an axom reducer is a better fit for what the code is trying to do here.  I'm OK merging in what you have here to fix the build issue and I can do a follow-on PR that uses `axom::ReduceSum` here instead.

ebchin (approved):

kennyweiss (approved):

PR #207, opened 2026-05-20
Energy mortar cleanup

ebchin:
- Removed .clangd as it was build specific
- Clean up coupling scheme and mfem_tribol logic
- Clean up tests (single registration of coupling scheme instead of per-timestep, remove formulation_ready in tribol_energy_mortar_patch_LM). Needed a method to update penalty parameters in ContactFormulation to support this.
- Add a getMfemJacobian() method to return a single mfem::HypreParMatrix (no blocks) for penalty methods


src/tests/tribol_energy_mortar_patch.cpp
```diff
@@ -172,6 +172,12 @@ class MfemMortarEnergyPatchTest : public testing::TestWithParam<std::tuple<int>>
     tribol::RealT dt = 1.0 / num_timesteps_;
     int cs_id = 0, mesh1_id = 0, mesh2_id = 1;
 
+    tribol::registerMfemCouplingScheme( cs_id, mesh1_id, mesh2_id, mesh, coords, mortar_attrs, nonmortar_attrs,
+                                        tribol::SURFACE_TO_SURFACE, tribol::NO_SLIDING, tribol::ENERGY_MORTAR,
+                                        tribol::FRICTIONLESS, tribol::PENALTY, tribol::BINNING_GRID );
```
srwopschall, line 177: This looks like it changed from tribol::LAGRANGE_MULTIPLIER to tribol::PENALTY. Please clarify.

Add a warning if someone specifies Lagrange multipliers with energy mortar.

src/tests/tribol_energy_mortar_patch.cpp
```diff
@@ -172,6 +172,12 @@ class MfemMortarEnergyPatchTest : public testing::TestWithParam<std::tuple<int>>
...
+    tribol::setLagrangeMultiplierOptions( cs_id, tribol::ImplicitEvalMode::MORTAR_RESIDUAL_JACOBIAN );
```
srwopschall: Consider: 

1) hard code Jacobian stuff that is necessary for energy mortar; a user shouldn't have to  call an API function to make required behavior possible.
2) Merge functionality in something like setResidualJacobianOptions() which specifies residual only eval, residual + global jacobian contributions etc. for any method. Can have sanity check for common plane etc.

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -22,6 +24,50 @@ MeshBuilder MeshBuilder::SquareMesh( int n_x_els, int n_y_els )
   return mfem::Mesh::MakeCartesian2D( n_x_els, n_y_els, mfem::Element::QUADRILATERAL );
 }
 
+MeshBuilder MeshBuilder::CircularShell( int n_radial_els, int n_hoop_els, double inner_radius, double outer_radius )
```
srwopschall: Call Cylinder2D

src/tests/tribol_energy_mortar_patch.cpp
```diff
@@ -197,25 +203,14 @@ class MfemMortarEnergyPatchTest : public testing::TestWithParam<std::tuple<int>>
...
 
-      auto A_cont_ptr = tribol::getMfemDfDx( cs_id );
+      auto A_cont_ptr = tribol::getMfemJacobian( cs_id );
       ASSERT_TRUE( A_cont_ptr != nullptr );
       shared::ParSparseMat A_cont( std::move( A_cont_ptr ) );
 
-      mfem::Vector f_contact( par_fe_space.GetTrueVSize() );
-      f_contact = 0.0;
-      tribol::getMfemResponse( cs_id, f_contact );
+      auto f_contact = tribol::getMfemTDofForce( cs_id );
```
srwopschall: I liked `getMfemResponse()` better from before. Do we have to use Mfem language and call it a TDof force? It makes it sound like there is some other force one could want or need too, which could be confusing. I think for the API the contact force response is just the contact force reponse. 

If we get rid of getMfemResponse() which returns an Ldof force, because we really want the Tdof force, then we change this routine to something like getMfemContactForce() and mention in the header it is Tdof.

src/tests/tribol_energy_mortar_patch_LM.cpp
```diff
@@ -237,14 +236,13 @@ class MfemMortarEnergyLagrangePatchTest : public testing::TestWithParam<std::tup
...
 
         tribol::updateMfemParallelDecomposition();
-        if ( formulation_ready ) {
-          // Set lambda for LM assembly prior to calling update().
-          auto& tribol_lambda = tribol::getMfemTDofPressure( cs_id );
-          tribol_lambda = 0.0;
-          tribol_lambda.Add( 1.0, lambda );
-        }
+
+        // Set lambda for LM assembly prior to calling update().
```
srwopschall, line 240: Clarify removing the guard checking if "formulation_ready". What did that mean originally anyway?

src/tests/tribol_energy_mortar_patch_LM.cpp
```diff
@@ -237,14 +236,13 @@ class MfemMortarEnergyLagrangePatchTest : public testing::TestWithParam<std::tup
...
+        auto& tribol_lambda = tribol::getMfemTDofPressure( cs_id );
```
srwopschall: Same comment about using TDof in the name.

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -1009,6 +1009,73 @@ INSTANTIATE_TEST_SUITE_P( MfemLorTransfer, MfemLorTransferParamTest,
                                            LorTransferParams{ 3, 4, FieldKind::LagrangeMultiplier } ),
                           ParamsToString );
 
+#ifdef TRIBOL_USE_ENZYME
+
+TEST_F( MfemJacobianTest, mfem_penalty_jacobian_retrieval )
+{
+  int n_ranks;
```
srwopschall, line 1016: Add a high level comment as to what this test is doing

src/tests/tribol_mfem_jacobian.cpp
```diff
@@ -1009,6 +1009,73 @@ INSTANTIATE_TEST_SUITE_P( MfemLorTransfer, MfemLorTransferParamTest,
...
+  tribol::setMfemKinematicConstantPenalty( cs_id, 1.0, 1.0 );
+
+  // Build internal MfemData
+  tribol::updateMfemParallelDecomposition();
+
+  double dt = 1.0;
+  tribol::update( 1, 1.0, dt );
+
+  // Retrieve Jacobian using the new penalty method
+  auto DfDx = tribol::getMfemJacobian( cs_id );
```
srwopschall: I'm noticing that this was getMfemDfDx() or something like that before, which is specific to a particular block, but now it just is getMfemJacobian(). Can you clarify whether this is sufficiently general, or is block specific, or what changed? Penalty enforcement should just have a DfDx block right? Whereas LM enforcement will have contributions to other blocks. Is this routine for both and returns the full matrix of ALL Jacobian contributions?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -292,10 +293,9 @@ void registerMfemCouplingScheme( IndexT cs_id, int mesh_id_1, int mesh_id_2, con
   auto& cs = CouplingSchemeManager::getInstance().at( cs_id );
   cs.setMPIComm( mesh.GetComm() );
 
-  // Set data required for use with Lagrange multiplier enforcement option.
-  // Coupling scheme validity will be checked later, but here some initial
-  // data is created/initialized for use with LMs.
-  if ( enforcement_method == LAGRANGE_MULTIPLIER || contact_method == ENERGY_MORTAR ) {
+  // Initialize methods that require definition of a submesh. Coupling scheme validity will be checked
```
srwopschall, line 301: maybe clarify that this initializes methods that require definition of field on a submesh.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -320,18 +320,21 @@ void registerMfemCouplingScheme( IndexT cs_id, int mesh_id_1, int mesh_id_2, con
...
              lm_options.eval_mode == ImplicitEvalMode::MORTAR_RESIDUAL_JACOBIAN ) ) ||
          contact_method == ENERGY_MORTAR ) {
-      // create matrix transfer operator between redecomp and
-      // parent/parent-linked boundary submesh
+      // create matrix transfer operator between redecomp and parent/parent-linked boundary submesh
       cs.setMfemJacobianData( std::make_unique<MfemJacobianData>( *mfem_data, *cs.getMfemSubmeshData() ) );
     }
   }
   cs.setMfemMeshData( std::move( mfem_data ) );
+  if ( contact_method == ENERGY_MORTAR ) {
```
srwopschall, line 340: Consider a note here that for now only ENERGY_MORTAR works with the contact formulation code path?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -382,6 +385,10 @@ void setMfemKinematicConstantPenalty( IndexT cs_id, RealT mesh1_penalty, RealT m
   cs->getMfemMeshData()->ClearAllPenaltyData();
   cs->getMfemMeshData()->SetMesh1KinematicConstantPenalty( mesh1_penalty );
   cs->getMfemMeshData()->SetMesh2KinematicConstantPenalty( mesh2_penalty );
+
+  if ( cs->hasContactFormulation() ) {
+    cs->getContactFormulation()->updatePenaltyParameters( true, mesh1_penalty );
```
srwopschall: To be consistent with other locations in the code we average both penalties. Consider passing both into updatePenaltyParameters() and then average them. 

Can you clarify what this routine is doing? It only takes the mesh1_penalty so I'm wondering what actually is being done.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -541,19 +548,9 @@ void getMfemResponse( IndexT cs_id, mfem::Vector& r )
...
-    const auto& f = cs->getContactFormulation()->getMfemForce();
-    if ( r.Size() == 0 ) {
-      r.SetSize( f.Size() );
-    }
-    SLIC_ERROR_ROOT_IF( r.Size() != f.Size(), "getMfemResponse(): size mismatch with formulation force vector." );
-    r = f;
-    return;
-  }
-
+  SLIC_ERROR_ROOT_IF(
```
srwopschall, line 556: I think it is confusing to have a getMfemResponse(), but then a getMfemTdofForce() if you use the contact formulation stuff. Can we unify?

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -567,7 +564,9 @@ mfem::HypreParVector getMfemTDofForce( IndexT cs_id )
       !cs, axom::fmt::format( "Coupling scheme cs_id={0} does not exist. Call tribol::registerMfemCouplingScheme() "
                               "to create a coupling scheme with this cs_id.",
                               cs_id ) );
-  SLIC_ERROR_ROOT_IF( !cs->hasContactFormulation(), "Coupling scheme does not contain a contact formulation." );
+  SLIC_ERROR_ROOT_IF(
+      !cs->hasContactFormulation(),
```
srwopschall, line 573: Again this is sort of confusing.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -579,6 +578,10 @@ std::unique_ptr<mfem::BlockOperator> getMfemBlockJacobian( IndexT cs_id )
                               "to create a coupling scheme with this cs_id.",
                               cs_id ) );
 
+  SLIC_ERROR_ROOT_IF( cs->getEnforcementMethod() == PENALTY,
+                      "getMfemBlockJacobian() is not supported for coupling schemes with penalty enforcement. "
+                      "Use getMfemJacobian() instead." );
```
srwopschall: Potential confusion.

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -627,22 +630,50 @@ std::unique_ptr<mfem::BlockOperator> getMfemBlockJacobian( IndexT cs_id )
...
+      dfdx->SetBlock( 1, 0, block_10.release() );
+    }
 
     return dfdx;
   } else {
     return BuildMfemBlockJacobian( *cs, *cs->getMethodData(), all_info, all_info );
   }
 }
 
+std::unique_ptr<mfem::HypreParMatrix> getMfemJacobian( IndexT cs_id )
```
srwopschall: Calling which routine where for if contact formulation or not is very confusing. Also, getMfemJacobian() makes it sound like it returns the entire Jacobian, regardless of enforcement procedure, where this actually only calls getMfemDfDx(). Looks like this is only for Penalty enforcement where the whole Jacobian is just DfDx? In this case, let's change the name of the routine.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1029,14 +1029,14 @@ void CouplingScheme::performBinning()
 //------------------------------------------------------------------------------
 int CouplingScheme::apply( int cycle, RealT t, RealT& dt )
 {
-  if ( m_formulation_impl ) {
-    if ( m_interface_pairs.size() > 0 ) {
-      m_formulation_impl->setInterfacePairs( std::move( m_interface_pairs ), 0 );
+  if ( m_formulation ) {
+    if ( m_interface_pairs.size() > 0 || !hasFixedBinning() ) {
```
srwopschall, line 1033: I'll have to remind myself what hasFixedBinning() is going to do in terms of getting into this conditional block.

src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1158,29 +1158,18 @@ int CouplingScheme::apply( int cycle, RealT t, RealT& dt )
...
-    }
-    if ( m_formulation_impl ) {
-      m_formulation_impl->updateMeshes( *m_mesh1, *m_mesh2 );
-    }
-  }
-
-  if ( m_formulation_impl ) {
-    this->m_isValid = true;
+    // pass the new meshes to the formulation
+    m_formulation->updateMeshes( *m_mesh1, *m_mesh2 );
```
srwopschall, line 1171: Presumably the contact formulation is now created elsewhere? (see line 1173 in diff)

src/tribol/mesh/MfemData.hpp
```diff
@@ -1790,10 +1790,10 @@ struct PackedPairJacobianContribs {
   const Array1D<int>* row_elem_map{ nullptr };
   /// Tribol element-id -> redecomp element-id map for column element ids
   const Array1D<int>* col_elem_map{ nullptr };
-  axom::Array<int> row_elem_ids;      ///< Tribol element IDs for rows
-  axom::Array<int> col_elem_ids;      ///< Tribol element IDs for columns
-  axom::Array<double> jacobian_data;  ///< Flattened Jacobian data
-  axom::Array<int> value_offsets;     ///< Offsets into jacobian_data for each element
+  Array1D<int, MemorySpace::Host> row_elem_ids;      ///< Tribol element IDs for rows
```
srwopschall, line 1793: Just assuming these were axom::Array s that weren't moved over to Array1D in a previous MR?

src/tribol/physics/ContactFormulation.hpp
```diff
@@ -111,6 +111,13 @@ class ContactFormulation {
    */
   virtual void updateMeshes( MeshData& /*mesh1*/, MeshData& /*mesh2*/ ) {}
 
+  /**
+   * @brief Update penalty parameters
+   *
+   * Used to dynamically update penalty stiffness after the formulation has been instantiated.
+   */
+  virtual void updatePenaltyParameters( bool /*use_penalty*/, double /*k*/ ) {}
```
srwopschall: For consistency with other parts of the code, take in both constant penalty stiffnesses, one for each mesh, and average them. 

Penalty Parameters is a little vague. This just sets the penalty stiffness and is only used for the constant penalty stiffness option. Consider updateConstantPenaltyStiffness() or something like that?

src/tribol/physics/ContactFormulationFactory.cpp
```diff
@@ -43,8 +43,8 @@ std::unique_ptr<ContactFormulation> createContactFormulation( CouplingScheme* cs
     SLIC_ERROR_ROOT_IF( !cs->hasMfemJacobianData(), "ENERGY_MORTAR requires MFEM Jacobian data." );
 
     return std::make_unique<EnergyMortarAdapter>( *cs->getMfemMeshData(), *cs->getMfemSubmeshData(),
```
srwopschall, line 46: Explain no longer needing the meshes?

src/shared/mesh/MeshBuilder.cpp
```diff
@@ -22,6 +24,50 @@ MeshBuilder MeshBuilder::SquareMesh( int n_x_els, int n_y_els )
...
+    double r = inner_radius + x * ( outer_radius - inner_radius );
+    double theta = y * 2.0 * pi;
+
+    nodes( vdof_x ) = r * std::cos( theta );
+    nodes( vdof_y ) = r * std::sin( theta );
+  }
+  return mesh;
+}
+
+MeshBuilder MeshBuilder::SemiCircularShell( int n_radial_els, int n_hoop_els, double inner_radius, double outer_radius )
```
srwopschall: HalfCylinder2D()

src/tribol/interface/mfem_tribol.cpp
```diff
@@ -627,22 +630,50 @@ std::unique_ptr<mfem::BlockOperator> getMfemBlockJacobian( IndexT cs_id )
...
+      !cs, axom::fmt::format( "Coupling scheme cs_id={0} does not exist. Call tribol::registerMfemCouplingScheme() "
+                              "to create a coupling scheme with this cs_id.",
+                              cs_id ) );
+
+  SLIC_ERROR_ROOT_IF( cs->getEnforcementMethod() != PENALTY,
+                      "getMfemJacobian() is only supported for coupling schemes with penalty enforcement. "
+                      "Use getMfemBlockJacobian() instead." );
+
+  if ( cs->hasContactFormulation() ) {
+    return cs->getContactFormulation()->getMfemDfDx();
```
srwopschall: Get rid of getMfemJacobian() API function and move this if block to the original getMfemDfDx() API function below. There may still be some circular logic with the contact formulation's getMfemDfDx() implementation but I'll let you sort that out.

srwopschall (commented):
@ebchin - We discussed most everything I wanted to discuss. The rest of the comments are self-explanatory in terms of what might need to be modified.

src/tests/tribol_energy_mortar_patch.cpp
ebchin: Added

src/tests/tribol_energy_mortar_patch.cpp
ebchin: `registerMfemCouplingScheme()` already creates MFEM Jacobian transfer data so this isn't needed

src/tests/tribol_energy_mortar_patch.cpp
ebchin: Now `getMfemContactForce()`

src/tests/tribol_energy_mortar_patch_LM.cpp
ebchin: The guard was only needed when formulation setup was not available at the right time. The current path creates the Energy Mortar contact formulation during MFEM coupling scheme registration, before the Newton loop accesses the pressure/multiplier vector.

src/tests/tribol_mfem_jacobian.cpp
ebchin: done

src/tests/tribol_mfem_jacobian.cpp
ebchin: back to getMfemDfDx(). removed getMfemJacobian()

src/tribol/interface/mfem_tribol.cpp
ebchin: The existing comment already says this initializes methods requiring a submesh definition and notes the LM setup use. I left this as-is.

src/tribol/interface/mfem_tribol.cpp
ebchin: I think this is clear enough from looking at the formulation code (e.g. the factory function).

src/tribol/interface/mfem_tribol.cpp
ebchin: Renamed the hook to `updateConstantPenaltyStiffness(mesh1_penalty, mesh2_penalty)` and average the two stiffnesses inside the Energy Mortar adapter. The factory also averages both stored MFEM constant penalties when initializing the formulation.

src/tribol/interface/mfem_tribol.cpp
ebchin: Renamed the contact formulation accessor to `getMfemContactForce()`

src/tribol/interface/mfem_tribol.cpp
ebchin: Removed

src/tribol/mesh/CouplingScheme.cpp
ebchin: the contact formulation is created during `registerMfemCouplingScheme()` after MFEM mesh, submesh, and Jacobian data are available

src/tribol/mesh/MfemData.hpp
ebchin: Right

src/tribol/physics/ContactFormulationFactory.cpp
ebchin: mesh handles are now refreshed through `updateMeshes()`, since redecomp will make the handles invalid.

src/tests/tribol_mfem_jacobian.cpp
srwopschall: Look at MFEM Api, it still looks like there is a getMfemJacobian(). Can you resolve this?

src/tribol/interface/mfem_tribol.cpp
srwopschall: Please align mfem api function level for getMfemDfDx() with contact formulation and verify there is no getMfemJacobian() routines floating around.

srwopschall (approved):
Resolve the getMfemJacobian() vs. getMfemDfDx() comments.

src/tribol/interface/mfem_tribol.cpp
ebchin: It was removed in the previous commit, this is just an outdated block. There are also already getMfemDfDx() etc. functions implemented in mfem_tribol.hpp/cpp

src/tests/tribol_mfem_jacobian.cpp
ebchin: It was removed, this is just an outdated view.

PR #208, opened 2026-05-28
Common plane higher order quadrature

ebchin:


src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -185,10 +185,139 @@ class CommonPlaneTest : public ::testing::Test {
...
+  RealT gap{ 0. };
+  RealT total_abs_force{ 0. };
+  RealT total_force_z{ 0. };
+};
+
+WarpedQuadForceResult runWarpedQuadForceCase( tribol::PolyInteg rule )
+{
+  constexpr int numVerts = 4;
+
+  RealT x1[numVerts] = { 0.0, 1.0, 1.0, 0.0 };
```
srwopschall, line 203: Minor: considering adding a comment about the configuration of the two blocks including the gap.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -185,10 +185,139 @@ class CommonPlaneTest : public ::testing::Test {
...
+
+  tribol::setKinematicConstantPenalty( 0, 1.0 );
+  tribol::setKinematicConstantPenalty( 1, 1.0 );
+
+  tribol::registerCouplingScheme( 0, 0, 1, tribol::SURFACE_TO_SURFACE, tribol::NO_CASE, tribol::COMMON_PLANE,
+                                  tribol::FRICTIONLESS, tribol::PENALTY, tribol::BINNING_GRID,
+                                  tribol::ExecutionMode::Sequential );
+
+  tribol::setPenaltyOptions( 0, tribol::KINEMATIC, tribol::KINEMATIC_CONSTANT );
+  tribol::setCommonPlaneIntegrationOptions( 0, rule, 3 );
```
srwopschall: Consider clarifying what 0 and 3 are since someone may come to the tests to see how this is used.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -185,10 +185,139 @@ class CommonPlaneTest : public ::testing::Test {
...
+    result.total_abs_force += std::abs( fx1[i] ) + std::abs( fy1[i] ) + std::abs( fz1[i] );
+    result.total_abs_force += std::abs( fx2[i] ) + std::abs( fy2[i] ) + std::abs( fz2[i] );
+    result.total_force_z += fz1[i] + fz2[i];
+  }
+
+  tribol::finalize();
+  return result;
+}
+
+struct EdgeLocalContactResult {
```
srwopschall: Is this struct intended to hold a force result. Consider consistent naming similar to the WarpedQuadForceResult.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -185,10 +185,139 @@ class CommonPlaneTest : public ::testing::Test {
...
+  RealT gap{ 0. };
+  RealT total_abs_force{ 0. };
+  RealT mesh1_node_force_norm[2] = { 0., 0. };
 };
 
+EdgeLocalContactResult runEdgeLocalContactCase( tribol::PolyInteg rule )
+{
+  constexpr int numVerts = 2;
+
+  RealT x1[numVerts] = { 1.0, 0.0 };
```
srwopschall, line 274: Minor: consider adding comment about configuration of the two edges. Is this an interpen interaction or full overlap?

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -185,10 +185,139 @@ class CommonPlaneTest : public ::testing::Test {
...
+
+  tribol::setKinematicConstantPenalty( 0, 1. );
+  tribol::setKinematicConstantPenalty( 1, 1. );
+
+  tribol::registerCouplingScheme( 0, 0, 1, tribol::SURFACE_TO_SURFACE, tribol::NO_CASE, tribol::COMMON_PLANE,
+                                  tribol::FRICTIONLESS, tribol::PENALTY, tribol::BINNING_GRID,
+                                  tribol::ExecutionMode::Sequential );
+
+  tribol::setPenaltyOptions( 0, tribol::KINEMATIC, tribol::KINEMATIC_CONSTANT );
+  tribol::setCommonPlaneIntegrationOptions( 0, rule, 4 );
```
srwopschall: Clarify what 0 and 4 are.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
   tribol::finalize();
 }
 
+TEST_F( CommonPlaneTest, multipoint_quad_execution )
+{
+  this->m_mesh.mortarMeshId = 0;
```
srwopschall, line 389: Consider adding comment about what this test is testing.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+  this->m_mesh.nonmortarMeshId = 1;
+
+  const int nElems = 2;
+  const RealT x_min1 = 0.;
```
srwopschall, line 393: Minor: consider comment about mesh configuration e.g. full overlap, interpen overlap, warped faces, planar faces etc.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+  EXPECT_EQ( err, 0 );
+
+  tribol::CouplingScheme* couplingScheme = &tribol::CouplingSchemeManager::getInstance().at( 0 );
+  compareGaps( couplingScheme, z_min2 - z_max1, 1.E-8, "kinematic_penetration" );
+  checkForceSense( couplingScheme );
+
+  tribol::finalize();
+}
+
+TEST_F( CommonPlaneTest, multipoint_triangle_execution )
```
srwopschall, line 428: Comment about test. Is this for linear triangles in contact or is this a triangulation of the overlap?

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+{
+  this->m_mesh.mortarMeshId = 0;
+  this->m_mesh.nonmortarMeshId = 1;
+
+  const int nElems = 2;
```
srwopschall, line 435: Minor: Consider comment about contact mesh configuration, e.g. interpen overlap, full overlap, warped faces, planar faces etc.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+
+TEST_F( CommonPlaneTest, multipoint_warped_quad_local_contact )
+{
+  const auto single_point = runWarpedQuadForceCase( tribol::SINGLE_POINT );
+  const auto multi_point = runWarpedQuadForceCase( tribol::MULTI_POINT );
+
+  EXPECT_EQ( single_point.err, 0 );
+  EXPECT_EQ( multi_point.err, 0 );
+
+  EXPECT_GT( single_point.gap, 0. );
```
srwopschall: I'm wondering can we actually check against the real gap, especially for the warped case. Even if there was a bug in the code, this `EXPECT_GT` will likely always be satisfied.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+  EXPECT_GT( multi_point.gap, 0. );
+
+  EXPECT_NEAR( single_point.total_abs_force, 0., 1.e-12 );
```
srwopschall, line 485: Can you clarify what we expect the forces to be. I am still concerned that these checks aren't going to catch any potential bugs in the quadrature especially in light of some of the projection nuances.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+  const auto single_point = runEdgeLocalContactCase( tribol::SINGLE_POINT );
+  const auto multi_point = runEdgeLocalContactCase( tribol::MULTI_POINT );
+
+  EXPECT_EQ( single_point.err, 0 );
+  EXPECT_EQ( multi_point.err, 0 );
+
+  EXPECT_EQ( single_point.num_active_pairs, 1 );
+  EXPECT_EQ( multi_point.num_active_pairs, 1 );
+
+  EXPECT_LT( single_point.gap, 0. );
```
srwopschall: Same as previous test. I'm concerned that these checks are too "soft" and may not catch any bugs especially with some of the projection nuances etc.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+  EXPECT_LT( multi_point.gap, 0. );
+
+  EXPECT_GT( single_point.total_abs_force, 1.e-6 );
+  EXPECT_GT( multi_point.total_abs_force, 1.e-6 );
+
+  const RealT single_point_imbalance =
```
srwopschall, line 511: Add comment about what an "imbalance" means here and what this is checking.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -245,6 +374,130 @@ TEST_F( CommonPlaneTest, penetration_gap_check )
...
+      std::abs( single_point.mesh1_node_force_norm[0] - single_point.mesh1_node_force_norm[1] );
+  const RealT multi_point_imbalance =
+      std::abs( multi_point.mesh1_node_force_norm[0] - multi_point.mesh1_node_force_norm[1] );
+
+  EXPECT_LT( single_point_imbalance, 1.e-3 );
```
srwopschall: Clarify how you got 1.e-3 and then adding `single_point_imbalance + 1.e-3`.

src/tests/tribol_common_plane_penalty.cpp
```diff
@@ -671,11 +924,12 @@ TEST_F( CommonPlaneTest, common_plane_viscous_tangential_2d )
   // by the gap and then divided amongst the edge nodes
   RealT force_y = 0.5 * gap / numVerts;
   RealT force_x = visc_coeff * ( vx1[0] - vx2[0] ) / numVerts;
+  constexpr RealT tol = 5.e-5;
```
srwopschall, line 948: Can you clarify why this tolerance went from 1.e-10 to 5.e-5? This is a big change.

src/tests/tribol_enforcement_options.cpp
```diff
@@ -162,6 +162,27 @@ TEST_F( EnforcementOptionsTest, penalty_kinematic_constant_error )
   delete mesh;
 }
 
+TEST_F( EnforcementOptionsTest, common_plane_integration_options_are_stored )
+{
+  tribol::TestMesh* mesh = new tribol::TestMesh();
```
srwopschall, line 169: Minor: high level comment about what this is testing.

src/tests/tribol_enforcement_options.cpp
```diff
@@ -162,6 +162,27 @@ TEST_F( EnforcementOptionsTest, penalty_kinematic_constant_error )
...
+  tribol::setKinematicConstantPenalty( 0, penalty );
+  tribol::setKinematicConstantPenalty( 1, penalty );
+  tribol::setPenaltyOptions( 0, tribol::KINEMATIC, tribol::KINEMATIC_CONSTANT );
+  tribol::setCommonPlaneIntegrationOptions( 0, tribol::MULTI_POINT, 4 );
+
+  tribol::CouplingSchemeManager& csManager = tribol::CouplingSchemeManager::getInstance();
+  tribol::CouplingScheme* scheme = &csManager.at( 0 );
+
+  EXPECT_EQ( scheme->getEnforcementOptions().penalty_options.common_plane_rule, tribol::MULTI_POINT );
+  EXPECT_EQ( scheme->getEnforcementOptions().penalty_options.common_plane_quadrature_order, 4 );
```
srwopschall: Formatting looks off here?

src/tests/tribol_iso_integ.cpp
```diff
@@ -18,6 +18,37 @@
...
+RealT factorial( int n )
+{
+  RealT result = 1.;
+  for ( int i = 2; i <= n; ++i ) {
+    result *= i;
+  }
+  return result;
+}
+
+RealT referenceTriangleMoment( int px, int py ) { return factorial( px ) * factorial( py ) / factorial( px + py + 2 ); }
```
srwopschall, line 32: Add comment. Which moment?

src/tests/tribol_iso_integ.cpp
```diff
@@ -18,6 +18,37 @@
...
+
+RealT normalizedReferenceTriangleMoment( int px, int py ) { return 2. * referenceTriangleMoment( px, py ); }
+
+RealT evalTriangleRuleMoment( bool use_legacy, int order, int px, int py )
```
srwopschall, line 39: Add comments about what these routines are doing including input/output? Someone may want to or need to use them in tests in the future?

src/tests/tribol_iso_integ.cpp
```diff
@@ -242,6 +273,40 @@ TEST_F( IsoIntegTest, nonaffine )
   EXPECT_EQ( convrg, true );
 }
 
+TEST( TriangleRuleTest, legacy_and_symmetric_match_on_shared_orders )
+{
+  for ( int order : { 2, 4 } ) {
```
srwopschall, line 283: Minor: add simple comment about what this is testing and what the expectation is.

src/tests/tribol_iso_integ.cpp
```diff
@@ -242,6 +273,40 @@ TEST_F( IsoIntegTest, nonaffine )
...
+    EXPECT_NEAR( evalTriangleRuleMoment( true, order, 0, 0 ), evalTriangleRuleMoment( false, order, 0, 0 ), 2.e-10 );
+    EXPECT_NEAR( evalTriangleRuleMoment( true, order, 2, 0 ), evalTriangleRuleMoment( false, order, 2, 0 ), 2.e-10 );
+    EXPECT_NEAR( evalTriangleRuleMoment( true, order, 1, 1 ), evalTriangleRuleMoment( false, order, 1, 1 ), 2.e-10 );
+  }
+}
+
+TEST( TriangleRuleTest, gauss_poly_int_tri_supports_order_10 )
+{
+  constexpr int dim = 3;
```
srwopschall, line 294: Minor: Comment. Is this checking to see if a gauss integration rule on triangle facets belonging to an overlap polygon in 3D is supported up to order 10?

src/tests/tribol_iso_integ.cpp
```diff
@@ -242,6 +273,40 @@ TEST_F( IsoIntegTest, nonaffine )
...
+  constexpr int num_nodes = 3;
+  RealT xyz[dim * num_nodes] = { 0., 0., 0., 1., 0., 0., 0., 1., 0. };
+
+  tribol::SurfaceContactElem elem( dim, xyz, xyz, xyz, num_nodes, num_nodes, nullptr, nullptr, 0, 0 );
+  tribol::IntegPts integ;
+  tribol::GaussPolyIntTri( elem, integ, 10 );
+
+  RealT area = 0.;
+  RealT moment73 = 0.;
```
srwopschall, line 303: Add comment about what moment73 means and what this is computing.

src/tests/tribol_mfem_common_plane.cpp
```diff
@@ -37,10 +37,12 @@
  * difference explicit time integration scheme.
  *
  * Both the element penalty and a constant penalty are tested, with the constant penalty tuned to match the element
- * penalty for this case.  As a result, the test comparisons are the same for both penalty types.
+ * penalty for this case. The contact solve is also exercised with both the legacy single-point CommonPlane integration
+ * rule and the new full triangle-decomposition rule.
```
srwopschall: Consider mentioning the order used for the full triangle-decomp rule. I recall this being a test(s) with planar contact faces so we don't expect to see a difference in the force or gap between single point integration and multipoint?

src/tribol/common/Parameters.hpp
```diff
@@ -235,8 +235,8 @@ enum IntNodalFields
  */
 enum PolyInteg
 {
-  SINGLE_POINT,     ///! Single point integration at centroid of polygon
-  FULL_TRI_DECOMP,  ///! Full integration using triangular decomposition
+  SINGLE_POINT,  ///! Single point integration at centroid of polygon
+  MULTI_POINT,   ///! Multi-point integration over the overlap
```
srwopschall: Does this always use a triangle decomposition of the overlap? Consider adding this in the doxygen if this is the case.

src/tribol/common/Parameters.hpp
```diff
@@ -444,6 +444,8 @@ struct PenaltyEnforcementOptions {
   PenaltyConstraintType constraint_type;
   KinematicPenaltyCalculation kinematic_calculation;
   RatePenaltyCalculation rate_calculation;
+  PolyInteg common_plane_rule{ SINGLE_POINT };
+  int common_plane_quadrature_order{ 3 };
```
srwopschall, line 449: If single_point is the default then I'm guessing `common_plane_quadrature_order` isn't used or doesn't really mean anything but having `SINGLE_POINT` be the default with a default order of 3 does look at odds with one another.

src/tribol/integ/FE.hpp
srwopschall, line 111: Since we have gone back and forth on this, let's add a note in this PR that the assumption is that x is a point in physical space ON the actual face/edge. In this note we should state that if it is not, then implicitly it should still work, but will ultimately find the (xi,eta) coordinate associated with the closest point to "x" on the face in the direction of the face normal. If that is actually happening, let's add this note.

src/tribol/integ/FE.hpp
```diff
@@ -437,6 +448,99 @@ TRIBOL_HOST_DEVICE inline int GetNumFaceNodes( int dim, FaceOrderType order_type
...
 }
 
+//-----------------------------------------------------------------------------
+/*!
+ *
+ * \brief Evaluates a linear basis function directly on a physical edge,
+ *        triangle, or quadrilateral face.
+ *
+ * \param [in] x pointer to stacked physical coordinates of the face vertices
+ * \param [in] pX x-coordinate of the evaluation point
```
srwopschall, line 463: Is the assumption that (pX,pY,pZ) is actually on the physical contact facet? If so, add to note.

src/tribol/integ/FE.hpp
```diff
@@ -437,6 +448,99 @@ TRIBOL_HOST_DEVICE inline int GetNumFaceNodes( int dim, FaceOrderType order_type
...
+ * \param [in] pY y-coordinate of the evaluation point
+ * \param [in] pZ z-coordinate of the evaluation point
```
srwopschall: Should this be set to zero for a physical edge? Consider adding note.

src/tribol/integ/FE.hpp
```diff
@@ -437,6 +448,99 @@ TRIBOL_HOST_DEVICE inline int GetNumFaceNodes( int dim, FaceOrderType order_type
...
+ * \param [in] numNodes number of nodes on the face
+ * \param [in] galerkinDim vector dimension of the nodal coefficients
+ * \param [in] nodeVals stacked nodal coefficients for the Galerkin approximation
+ * \param [in,out] galerkinVal evaluated Galerkin approximation
+ *
+ * \note This helper is topology-aware and supports the linear segment,
+ *       triangle, and quadrilateral basis evaluations used by CommonPlane.
+ *
+ */
+TRIBOL_HOST_DEVICE inline void GalerkinEvalOnPhysicalFace( const RealT* const x, const RealT pX, const RealT pY,
```
srwopschall, line 533: Sanity check. Are these new routines? The side-by-side code comparison isn't clear to me.

src/tribol/integ/FE.hpp
```diff
@@ -437,6 +448,99 @@ TRIBOL_HOST_DEVICE inline int GetNumFaceNodes( int dim, FaceOrderType order_type
...
+}
+
+//-----------------------------------------------------------------------------
+/*!
+ *
+ * \brief Evaluates a vector-valued Galerkin approximation directly on a physical
+ *        edge, triangle, or quadrilateral face.
+ *
+ * \param [in] x pointer to stacked physical coordinates of the face vertices
+ * \param [in] pX x-coordinate of the evaluation point
```
srwopschall, line 519: Add assumption if (pX, pY, pZ) is on the contact facet or not. Seems like it needs to lie on the facet since InvIso assumes it does, otherwise some bad/nuanced implicit behavior may happen.

src/tribol/integ/Integration.hpp
```diff
@@ -90,6 +90,18 @@ template <ContactMethod M, PolyInteg I>
 TRIBOL_HOST_DEVICE inline void EvalWeakFormIntegral( SurfaceContactElem const& elem, RealT* const integ1,
                                                      RealT* const integ2 );
 
+/// Selector for the triangle quadrature family used by GaussPolyIntTri().
+enum TriangleQuadratureRuleFamily
+{
+  TRI_RULE_LEGACY,
```
srwopschall, line 96: Please add comments as to what these are.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
 // Implementations
 //-----------------------------------------------------------------------------
 
-template <>
-TRIBOL_HOST_DEVICE inline void EvalWeakFormIntegral<COMMON_PLANE, SINGLE_POINT>( SurfaceContactElem const& elem,
-                                                                                 RealT* const integ1,
-                                                                                 RealT* const integ2 )
+TRIBOL_HOST_DEVICE inline void GetCommonPlaneOverlapCentroid( SurfaceContactElem const& elem, RealT cx[3] )
```
srwopschall, line 191: Doxygen?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
-    //    {
-    //       SLIC_DEBUG(elem.overlapCoords[elem.dim*i] << ", " << elem.overlapCoords[elem.dim*i+1]);
-    //    }
-    //    else
-    //    {
-    //       SLIC_DEBUG(elem.overlapCoords[elem.dim*i] << ", " << elem.overlapCoords[elem.dim*i+1] << ", "
-    //       elem.overlapCoords[elem.dim*i+2]);
-    //    }
-    // }
+TRIBOL_HOST_DEVICE inline void AccumulateCommonPlaneIntegralAtPoint( SurfaceContactElem const& elem, const RealT x[3],
```
srwopschall, line 204: Doxygen. What integral are we accumulating? This name is a little vague.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    RealT phi2 = 0.;
+    EvalBasisOnPhysicalFace( elem.faceCoords1, x[0], x[1], x[2], elem.numFaceVert, a, phi1 );
+    EvalBasisOnPhysicalFace( elem.faceCoords2, x[0], x[1], x[2], elem.numFaceVert, a, phi2 );
+    integ1[a] += wt * phi1;
+    integ2[a] += wt * phi2;
+  }
+}
+
+/*!
+ * \brief Returns the legacy triangle quadrature rule historically used by
```
srwopschall, line 219: Is the legacy rule just single point integration?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+ *        CommonPlane and GaussPolyIntTri().
+ *
+ * \param [in] order requested rule order
+ * \param [out] wts quadrature weights normalized so they sum to 1 on a triangle
+ * \param [out] coords quadrature coordinates stored as stacked (xi, eta) pairs
+ *
+ * \note This legacy rule is only available for the previously supported
+ *       orders 2 and 3/4 and is kept for regression comparison tests.
+ */
+TRIBOL_HOST_DEVICE inline int GetLegacyTriangleRule( int order, RealT* wts, RealT* coords )
```
srwopschall, line 229: Please clarify what this rule is.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    }
+    default:
+#ifdef TRIBOL_USE_HOST
+      SLIC_ERROR( "GetLegacyTriangleRule(): only legacy Gauss integration of order 2-4 is implemented." );
+#endif
+      return 0;
+  }
+}
+
+namespace detail {
```
srwopschall, line 281: What is the new namespace for? `detail` isn't very descriptive.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+
+/*!
+ * \brief Minimal symmetric triangle quadrature orbit data.
```
srwopschall, line 284: Not everyone is going to understand what "orbit" data is referring to. Can you clarify?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+ *       "On the identification of symmetric quadrature rules for finite element methods",
+ *       Computers & Mathematics with Applications 69(10), 2015,
+ *       doi:10.1016/j.camwa.2015.03.017.
+ *
+ *       PETSc stores weights for a reference triangle of area 2. Tribol uses
+ *       weights normalized so the weights sum to 1 and the physical triangle
+ *       area is applied separately, so the imported weights are scaled by 1/2
+ *       during expansion.
+ */
+struct SymmetricTriangleRuleData {
```
srwopschall, line 298: Can you clarify why this rule vs. some other rule? Did codex just choose to grab this from petsc? Or do you have specific justification as to why use this rule vs. expanding the legacy rule? Please clarify.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+  int num_centroid_orbits;
+  int num_edge_orbits;
+  int num_general_orbits;
+  const RealT* weights;
+  const RealT* orbits;
+};
+
+constexpr RealT symmetric_triangle_weight_scale = 0.5;
+
+constexpr RealT tri_deg2_weights[] = { 6.66666666666666666666666666666666635e-01 };
```
srwopschall, line 308: Sometimes codex etc. gets this stuff wrong, or flipped, or whatever. Do we have an explicit test that this integration rule has the correct numbers coded up? Or do we only know because of our contact tests that check force? Even still, those tests typically have planar faces etc., so I don't want anything to be missed.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+      return true;
+    case 10:
+      rule = { 1, 2, 3, tri_deg10_weights, tri_deg10_orbits };
+      return true;
+    default:
+      return false;
+  }
+}
+
+TRIBOL_HOST_DEVICE inline int ExpandSymmetricTriangleRule( const SymmetricTriangleRuleData& rule, RealT* wts,
```
srwopschall, line 414: Please add doxygen. What does it mean to expand the rule? What is the intent here?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+ * \brief Minimal symmetric triangle quadrature orbit data.
+ *
+ * \note The compact orbit tables below are adapted from the symmetric triangle
+ *       rules distributed in PETSc, which cite
+ *       F.D. Witherden and P.E. Vincent,
+ *       "On the identification of symmetric quadrature rules for finite element methods",
+ *       Computers & Mathematics with Applications 69(10), 2015,
+ *       doi:10.1016/j.camwa.2015.03.017.
+ *
+ *       PETSc stores weights for a reference triangle of area 2. Tribol uses
```
srwopschall, line 293: Did Tribol originally use this normalization? Is scaling the imported weights by 1/2 appropriate or appropriate in all cases when using this rule?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+}
+
+TRIBOL_HOST_DEVICE inline int ExpandSymmetricTriangleRule( const SymmetricTriangleRuleData& rule, RealT* wts,
+                                                           RealT* coords )
+{
+  int w_idx = 0;
+  int c_idx = 0;
+  int q_idx = 0;
+
+  for ( int orbit = 0; orbit < rule.num_centroid_orbits; ++orbit ) {
```
srwopschall, line 421: Might be useful to add comments to each little block of code to clarify what is happening.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    const RealT a = rule.orbits[c_idx++];
+    wts[q_idx] = symmetric_triangle_weight_scale * rule.weights[w_idx++];
+    coords[2 * q_idx] = a;
+    coords[2 * q_idx + 1] = a;
+    ++q_idx;
+  }
+
+  for ( int orbit = 0; orbit < rule.num_edge_orbits; ++orbit ) {
```
srwopschall, line 429: Minor: considering placing `int c_idx=0` right above the for loop that uses this as a counter.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    const RealT a = rule.orbits[c_idx++];
+    const RealT b = rule.orbits[c_idx++];
+    const RealT w = symmetric_triangle_weight_scale * rule.weights[w_idx++];
+
+    wts[q_idx] = w;
```
srwopschall, line 434: Does `q_idx` need to be reset to 0 prior to this second for loop?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    coords[2 * q_idx] = a;
+    coords[2 * q_idx + 1] = b;
+    ++q_idx;
+
+    wts[q_idx] = w;
```
srwopschall, line 439: swapping a and be etc in these little blocks is cryptic. I have no idea what is going on here.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    coords[2 * q_idx + 1] = a;
+    ++q_idx;
+
+    wts[q_idx] = w;
+    coords[2 * q_idx] = a;
+    coords[2 * q_idx + 1] = a;
+    ++q_idx;
+  }
+
+  for ( int orbit = 0; orbit < rule.num_general_orbits; ++orbit ) {
```
srwopschall, line 450: Does c_idx need to be reset to zero prior to entering this for loop?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    const RealT a = rule.orbits[c_idx++];
+    const RealT b = rule.orbits[c_idx++];
+    const RealT c = rule.orbits[c_idx++];
+    const RealT w = symmetric_triangle_weight_scale * rule.weights[w_idx++];
```
srwopschall, line 454: Same. Does w_idx need to be reset to 0? If the counters don't need to be reset we definitely need comments about how this accumulation is happening or what it is doing.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+
+    const RealT xi_eta[6][2] = { { b, c }, { c, b }, { a, c }, { c, a }, { a, b }, { b, a } };
```
srwopschall, line 456: Where does this come from?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    for ( int i = 0; i < 6; ++i ) {
```
srwopschall, line 457: Can you clarify what the hard coded "6" and "2" are?

src/tribol/integ/Integration.hpp
srwopschall, line 414: What is this routine returning?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+}  // namespace detail
+
+/*!
+ * \brief Returns the built-in higher-order symmetric triangle quadrature rule.
+ *
+ * \param [in] order requested rule order
+ * \param [out] wts quadrature weights normalized so they sum to 1 on a triangle
+ * \param [out] coords quadrature coordinates stored as stacked (xi, eta) pairs
+ *
+ * \note Orders 2 through 10 are available. Order 3 uses the same minimal rule
```
srwopschall, line 477: Clarify what the return integer means?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+ *       as order 4, matching the PETSc/Witherden-Vincent data set.
+ */
+TRIBOL_HOST_DEVICE inline int GetCommonPlaneTriangleRule( int order, RealT* wts, RealT* coords )
```
srwopschall, line 480: If this routine is only meant for the symmetric rule please not in the doxygen and consider renaming to something like GetSymmetricCommonPlaneTriangleRule() for clarity.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+#ifdef TRIBOL_USE_HOST
+    SLIC_ERROR( "GetCommonPlaneTriangleRule(): only symmetric triangle integration of order 2-10 is implemented." );
+#endif
+    return 0;
+  }
+  return detail::ExpandSymmetricTriangleRule( rule, wts, coords );
+}
+
+/*!
+ * \brief Returns the requested triangle quadrature rule family.
```
srwopschall, line 493: This brief makes it sound like it returns the value of the family enum. Consider clarifying.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+ *
+ * \param [in] order requested rule order
+ * \param [in] family selector for legacy versus symmetric rule data
+ * \param [out] wts quadrature weights normalized so they sum to 1 on a triangle
+ * \param [out] coords quadrature coordinates stored as stacked (xi, eta) pairs
+ */
+TRIBOL_HOST_DEVICE inline int GetTriangleRule( int order, TriangleQuadratureRuleFamily family, RealT* wts,
```
srwopschall, line 500: Clarify what the return integer means.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+}
+
+/*!
+ * \brief Returns a Gauss-Legendre quadrature rule on the unit segment.
+ *
+ * \param [in] order requested rule order
+ * \param [out] wts quadrature weights normalized so they sum to 1 on [0,1]
+ * \param [out] coords quadrature coordinates on [0,1]
+ */
+TRIBOL_HOST_DEVICE inline int GetCommonPlaneSegmentRule( int order, RealT* wts, RealT* coords )
```
srwopschall, line 523: Return integer meaning?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+  }
+}
+
+/*!
+ * \brief Returns a Gauss-Legendre quadrature rule on the unit segment.
+ *
+ * \param [in] order requested rule order
+ * \param [out] wts quadrature weights normalized so they sum to 1 on [0,1]
+ * \param [out] coords quadrature coordinates on [0,1]
+ */
```
srwopschall, line 522: Add a doxygen note about how high order this goes.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+      return 10;
+    default:
+#ifdef TRIBOL_USE_HOST
+      SLIC_ERROR( "GetCommonPlaneSegmentRule(): only Gauss-Legendre integration of order 2-10 is implemented." );
+#endif
+      return 0;
   }
+}
+
+TRIBOL_HOST_DEVICE inline void EvalWeakFormIntegralCommonPlaneMultiPoint( SurfaceContactElem const& elem,
```
srwopschall, line 660: I haven't seen it yet, but is there an EvalWeakFormIntegralCommonPlaneSinglePoint? Can we merge the two where you have one EvalWeakFormIntegral() used in common plane that handles the single point approach and the multipoint approach?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
-    //    }
-    // }
+TRIBOL_HOST_DEVICE inline void AccumulateCommonPlaneIntegralAtPoint( SurfaceContactElem const& elem, const RealT x[3],
+                                                                     const RealT wt, RealT* const integ1,
+                                                                     RealT* const integ2 )
+{
+  for ( int a = 0; a < elem.numFaceVert; ++a ) {
+    RealT phi1 = 0.;
+    RealT phi2 = 0.;
+    EvalBasisOnPhysicalFace( elem.faceCoords1, x[0], x[1], x[2], elem.numFaceVert, a, phi1 );
```
srwopschall, line 211: Can you confirm whether `elem.faceCoords1` are the projected planar face coordinates in 3D, or if they are the warped. If this PR switches them from planar to current configuration and possibly warped we need to check everywhere `elem.faceCoords1` etc. are used to see if this breaks any assumptions elsewhere.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+    const RealT x0 = elem.overlapCoords[0];
+    const RealT y0 = elem.overlapCoords[1];
+    const RealT x1 = elem.overlapCoords[2];
+    const RealT y1 = elem.overlapCoords[3];
+    const RealT length = magnitude( x1 - x0, y1 - y0 );
+
+    for ( int qp = 0; qp < num_qpts; ++qp ) {
+      const RealT s = rule_coords[qp];
+      const RealT one_minus_s = 1. - s;
+      RealT x[3] = { one_minus_s * x0 + s * x1, one_minus_s * y0 + s * y1, 0. };
```
srwopschall, line 678: Does `x` hold the segment vertices as mapped to the interval [0,1]? Or is it [-1,1]?

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+
+    const RealT area = Area3DTri( xTri, yTri, zTri );
+    if ( area <= 0. ) {
+      continue;
+    }
+
+    for ( int qp = 0; qp < num_qpts; ++qp ) {
+      const RealT xi = rule_coords[2 * qp];
+      const RealT eta = rule_coords[2 * qp + 1];
+      const RealT n0 = 1. - xi - eta;
```
srwopschall, line 716: Minor: add clarifying comment about forward map of (xi,eta) to x.

src/tribol/integ/Integration.hpp
```diff
@@ -174,92 +188,571 @@ int NumTWBPointsPerTri( int order );
...
+      RealT x[3];
+      x[0] = n0 * xTri[0] + xi * xTri[1] + eta * xTri[2];
+      x[1] = n0 * yTri[0] + xi * yTri[1] + eta * yTri[2];
+      x[2] = n0 * zTri[0] + xi * zTri[1] + eta * zTri[2];
+      AccumulateCommonPlaneIntegralAtPoint( elem, x, area * rule_wts[qp], integ1, integ2 );
```
srwopschall, line 721: Again, we need to clarify if the face coords in `elem` are warped, projected planar, or projected common plane coordinates.

src/tribol/interface/tribol.cpp
```diff
@@ -96,6 +96,28 @@ void setPenaltyOptions( IndexT cs_id, PenaltyConstraintType pen_enfrc_option,
...
+                               << "to calling this routine." );
+
+  auto& penalty_options = cs->getEnforcementOptions().penalty_options;
+
+  if ( !in_range( rule, NUM_INTEG_RULES ) ) {
+    SLIC_WARNING_ROOT( "tribol::setCommonPlaneIntegrationOptions(): polygon integration rule not available." );
+    return;
+  }
+
+  SLIC_ERROR_ROOT_IF( quadrature_order < 2 || quadrature_order > 10,
```
srwopschall, line 114: I think this API should handle all cases. I realize single point is the default, but this routine should be able to be "accidentally" called in order to handle single point and set things appropriately. Imagine in a host code that you'd have to guard this call against a quadrature order input value where if it is single point you can't call this and if it is higher than single then you do call this. We should just be able to call this no matter.

src/tribol/interface/tribol.cpp
```diff
@@ -96,6 +96,28 @@ void setPenaltyOptions( IndexT cs_id, PenaltyConstraintType pen_enfrc_option,
...
+                      "tribol::setCommonPlaneIntegrationOptions(): CommonPlane quadrature order must be in [2,10]." );
+
+  penalty_options.common_plane_rule = rule;
```
srwopschall, line 117: These really aren't penalty options. They are integration rule options for evaluation of the weak form integrals. Can we move this somewhere else? Happy to discuss where.

src/tribol/interface/tribol.hpp
```diff
@@ -65,6 +65,16 @@ void setPenaltyOptions( IndexT cs_id, PenaltyConstraintType pen_enfrc_option,
                         KinematicPenaltyCalculation kinematic_calc,
                         RatePenaltyCalculation rate_calc = NO_RATE_PENALTY );
 
+/*!
+ * \brief Sets CommonPlane overlap integration options
+ *
+ * \param [in] cs_id coupling scheme id
+ * \param [in] rule polygon integration rule for CommonPlane force integration
+ * \param [in] quadrature_order order of the CommonPlane quadrature used by MULTI_POINT in the range [2,10]
```
srwopschall, line 73: Fix doxygen if we make this general to support setting single point and multipoint options in this single routine.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
 
 namespace tribol {
 
-TRIBOL_HOST_DEVICE RealT ComputeGapRatePressure( CommonPlanePair& plane, const MeshData::Viewer& m1,
-                                                 const MeshData::Viewer& m2, RealT element_penalty,
-                                                 RatePenaltyCalculation rate_calc )
-{
-  auto fId1 = plane.getCpElementId1();
-  auto fId2 = plane.getCpElementId2();
+namespace {
```
srwopschall, line 20: Is this namespace new? Can you calrify? The side-by-side isn't super clear.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+             ( m1.getElementData().m_rate_percent_stiffness + m2.getElementData().m_rate_percent_stiffness );
     }
     default:
-      // no-op, quiet compiler
-      break;
+      return 0.;
+  }
+}
+
+TRIBOL_HOST_DEVICE inline bool Solve3x3( const RealT A[3][3], const RealT b[3], RealT x[3] )
```
srwopschall, line 45: Can we put this routine with the math helper routines?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+  x[0] = inv_detA * ( b[0] * ( A[1][1] * A[2][2] - A[1][2] * A[2][1] ) - A[0][1] * ( b[1] * A[2][2] - A[1][2] * b[2] ) +
+                      A[0][2] * ( b[1] * A[2][1] - A[1][1] * b[2] ) );
+  x[1] = inv_detA * ( A[0][0] * ( b[1] * A[2][2] - A[1][2] * b[2] ) - b[0] * ( A[1][0] * A[2][2] - A[1][2] * A[2][0] ) +
+                      A[0][2] * ( A[1][0] * b[2] - b[1] * A[2][0] ) );
+  x[2] = inv_detA * ( A[0][0] * ( A[1][1] * b[2] - b[1] * A[2][1] ) - A[0][1] * ( A[1][0] * b[2] - b[1] * A[2][0] ) +
+                      b[0] * ( A[1][0] * A[2][1] - A[1][1] * A[2][0] ) );
+  return true;
+}
+
+TRIBOL_HOST_DEVICE inline void AccumulateFaceInterpolation( const RealT* face_coords, const int num_nodes,
```
srwopschall, line 65: Doxygen? I still need to see where this is used but is this essentially the forward map of a (xi,eta) point? Except here you have the basis function `phi` already evaluated at that (xi,eta) and sent to this routine? Is it more useful to just pass in the (xi,eta) coords?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+                                                            const RealT* phi, RealT x_face[3], const int value_dim = 0,
+                                                            const RealT* nodal_vals = nullptr, RealT* values = nullptr )
+{
+  initRealArray( x_face, max_dim, 0. );
+  if ( values != nullptr ) {
```
srwopschall, line 70: Should we throw an error if values is a null pointer?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    initRealArray( values, value_dim, 0. );
+  }
+
+  for ( int a = 0; a < num_nodes; ++a ) {
+    x_face[0] += face_coords[3 * a] * phi[a];
```
srwopschall, line 75: Does this routine support 2D as well? Should there be a check?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    x_face[1] += face_coords[3 * a + 1] * phi[a];
+    x_face[2] += face_coords[3 * a + 2] * phi[a];
+
+    if ( values != nullptr ) {
+      for ( int i = 0; i < value_dim; ++i ) {
+        values[i] += nodal_vals[i + a * value_dim] * phi[a];
```
srwopschall, line 81: Is this a Galerkin evaluation using `nodal_vals` coefficients? We have a GalerkinEval() routine. Wondering if this is code duplication.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+      }
+    }
+  }
+}
+
+TRIBOL_HOST_DEVICE inline bool EvalLinearFaceAtProjectedPoint( const RealT* face_coords, const int num_nodes,
```
srwopschall, line 87: Doxygen? The name of this routine isn't clear to me what it is doing. Consider naming. This routine looks as if codex just wrote a ton of code. I got to the point where I just could not follow. We need to add some documentation as to what this is doing. If this takes a point that lies off the face and projects in the direction of `projection_dir` to the face and then needs the corresponding (xi,eta) coords in parent space then it seems like there is some code duplication with InvIso. Is that the case? If so, can we map out what unifying some of this machinery may look like? If this was codex generated then it just wouldn't know about this or is not directed to make use of it.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+                                                               const RealT x_query[3], const RealT projection_dir[3],
+                                                               RealT x_face[3], RealT* phi, const int value_dim = 0,
+                                                               const RealT* nodal_vals = nullptr,
+                                                               RealT* values = nullptr )
+{
+  // CommonPlane quadrature points lie on the overlap polygon. Evaluate the face
```
srwopschall, line 93: I wonder if this comment should go before wherever this routine is called. Here there is no way to check if `projected_dir` is in fact the common plane normal so this routine should maybe stay sufficiently general.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+  // fields at the corresponding on-face point found by projecting along the
+  // common-plane normal rather than by an off-surface closest-point inverse map.
+  initRealArray( phi, num_nodes, 0. );
+
+  if ( num_nodes == 3 ) {
+    const RealT x0[3] = { face_coords[0], face_coords[1], face_coords[2] };
+    const RealT e1[3] = { face_coords[3] - x0[0], face_coords[4] - x0[1], face_coords[5] - x0[2] };
```
srwopschall, line 100: Minor: add comment as to what e1 and e2 are. Are these in plane basis vectors in 3D space?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    const RealT e2[3] = { face_coords[6] - x0[0], face_coords[7] - x0[1], face_coords[8] - x0[2] };
+
+    RealT n_face[3];
+    crossProd( e1[0], e1[1], e1[2], e2[0], e2[1], e2[2], n_face[0], n_face[1], n_face[2] );
+
+    const RealT denom =
```
srwopschall, line 106: Add comments as to what the math is doing here. Is the assumption that the face is planar? Or can it be warped? If it is warped then the normal `n_face` will be different based on where it is evaluated. So should the bases vectors e1 and e2 be tangential to the face at the query point? Let's discuss the intent.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+        dotProd( n_face[0], n_face[1], n_face[2], projection_dir[0], projection_dir[1], projection_dir[2] );
+    constexpr RealT parallel_tol = 1.e-14;
+    if ( std::abs( denom ) <= parallel_tol ) {
+      return false;
+    }
+
+    const RealT dx0 = x0[0] - x_query[0];
```
srwopschall, line 113: Add comment. What is "step"?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    const RealT dy0 = x0[1] - x_query[1];
+    const RealT dz0 = x0[2] - x_query[2];
+    const RealT step = dotProd( n_face[0], n_face[1], n_face[2], dx0, dy0, dz0 ) / denom;
+
+    x_face[0] = x_query[0] + step * projection_dir[0];
+    x_face[1] = x_query[1] + step * projection_dir[1];
+    x_face[2] = x_query[2] + step * projection_dir[2];
+
+    const RealT r[3] = { x_face[0] - x0[0], x_face[1] - x0[1], x_face[2] - x0[2] };
```
srwopschall, line 122: Minor: consider adding comment as to what this block does.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    const RealT rhs1 = dotProd( e1[0], e1[1], e1[2], r[0], r[1], r[2] );
+    const RealT rhs2 = dotProd( e2[0], e2[1], e2[2], r[0], r[1], r[2] );
+
+    const RealT gram_det = gram11 * gram22 - gram12 * gram12;
+    constexpr RealT gram_tol = 1.e-15;
+    if ( std::abs( gram_det ) <= gram_tol ) {
+      return false;
+    }
+
+    const RealT xi = ( gram22 * rhs1 - gram12 * rhs2 ) / gram_det;
```
srwopschall, line 135: Add comment as to what this xi, eta calc does and what phi basis is being used.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+    phi[2] = eta;
+  } else if ( num_nodes == 4 ) {
+    constexpr int max_iter = 25;
+    constexpr RealT step_tol = 1.e-12;
+    constexpr RealT residual_tol = 1.e-12;
+    constexpr RealT xi_tol = 1.e-8;
+
+    RealT xi = 0.;
+    RealT eta = 0.;
+    RealT phi0[max_nodes_per_face] = { 0.25, 0.25, 0.25, 0.25 };
```
srwopschall, line 148: Clarify what this is and where it comes from .

src/tribol/physics/CommonPlane.cpp
```diff
@@ -17,48 +17,342 @@
...
+      for ( int i = 0; i < value_dim; ++i ) {
+        values[i] += nodal_vals[i + a * value_dim] * phi[a];
+      }
+    }
+  }
+
+  return true;
+}
+
+TRIBOL_HOST_DEVICE inline bool EvalLinearEdgeAtProjectedPoint( const RealT* edge_coords, const RealT x_query[2],
```
srwopschall, line 251: Doxygen and comments.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -77,14 +371,14 @@ TRIBOL_HOST_DEVICE RealT ComputeGapRatePressure( CommonPlanePair& plane, const M
   RealT cXf1 = plane.m_cXf1;
   RealT cYf1 = plane.m_cYf1;
   RealT cZf1 = ( dim == 3 ) ? plane.m_cZf1 : 0.;
-  GalerkinEval( x1, cXf1, cYf1, cZf1, LINEAR, PHYSICAL, dim, dim, v1, vel_f1 );
+  GalerkinEvalOnPhysicalFace( x1, cXf1, cYf1, cZf1, numNodesPerFace1, dim, v1, vel_f1 );
```
srwopschall, line 374: Per line 351 the routine getFace1Coords for the common plane "plane" object returns the average face-plane coordinates, not the current configuration and potentially warped coordinates. The use of the averaged face-plane coordinates for common plane is everywhere. By contrast, the mortar "plane" object uses the warped current configuration coords. These are all set on the `checkFacePair()` routine on each of the contact plane classes. If we modify how this is being handled for common plane we have to be certain we don't mess up any assumption elsewhere about which version of the face is being used. 

Consider renaming routine since it is not the current configuration physical face. Does `GalerkinEvalOnPhysicalFace` duplicate code in the original `GalerkinEval()` routine?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
     cntctElem.overlapNormal = overlapNormal;
     cntctElem.overlapArea = plane.m_area;
 
+    if ( use_multi_point ) {
```
srwopschall, line 584: Ah, I see that my comment above is all good if that is the single integration code path? Here we need to make sure we add a comment that we are explicitly choosing to use the current configuration physical face coordinates. This stuff is confusing and should be noted accordingly.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+      constexpr int max_qpts = max_symmetric_triangle_qpts;
+      RealT rule_wts[max_qpts] = { 0. };
+      RealT rule_coords[2 * max_qpts] = { 0. };
+      bool has_contact = false;
+
+      if ( dim == 3 ) {
+        const int num_qpts =
+            GetCommonPlaneTriangleRule( pen_enfrc_options.common_plane_quadrature_order, rule_wts, rule_coords );
+
+        RealT centroid[3];
```
srwopschall, line 611: Around this line and down is code duplication of code I saw in another routine. Can you look into consolidating?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+            x_q[2] = n0 * zTri[0] + xi * zTri[1] + eta * zTri[2];
+
+            RealT phi_q1[max_nodes_per_face] = { 0., 0., 0., 0. };
+            RealT phi_q2[max_nodes_per_face] = { 0., 0., 0., 0. };
+            RealT x_qf1[max_dim];
+            RealT x_qf2[max_dim];
+            RealT vel_q1[max_dim] = { 0., 0., 0. };
+            RealT vel_q2[max_dim] = { 0., 0., 0. };
+
+            const bool mapped_face1 = EvalLinearFaceAtProjectedPoint(
```
srwopschall, line 651: I'm wondering if we should add a unit test for two physical configuration faces and explicitly check the gap calculation at all the integration points for a representative higher order quadrature rule.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+                actual_xf1, num_nodes_per_face, x_q, overlapNormal, x_qf1, phi_q1, use_rate ? dim : 0,
+                use_rate ? &actual_vf1[0] : nullptr, use_rate ? vel_q1 : nullptr );
+            const bool mapped_face2 = EvalLinearFaceAtProjectedPoint(
+                actual_xf2, num_nodes_per_face, x_q, overlapNormal, x_qf2, phi_q2, use_rate ? dim : 0,
+                use_rate ? &actual_vf2[0] : nullptr, use_rate ? vel_q2 : nullptr );
+            if ( !mapped_face1 || !mapped_face2 ) {
```
srwopschall, line 657: Consider adding comment here and at the EvalLinearFaceAtProjectedPoint as to what a false return indicates. If we are in this if-block should there be a debug warning or something?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+              continue;
+            }
+
+            RealT local_gap = ( x_qf1[0] - x_qf2[0] ) * overlapNormal[0] + ( x_qf1[1] - x_qf2[1] ) * overlapNormal[1] +
```
srwopschall, line 661: Sanity check. Shouldn't x_qf1 - x_qf2 already be in the direction of the overlap normal?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+                              ( x_qf1[2] - x_qf2[2] ) * overlapNormal[2];
+            if ( local_gap > gap_tol ) {
```
srwopschall, line 663: Sanity check: just make sure the gap calculation of face1 - face2 is consistent with what was done before and the inequality in this gap check .

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+        const int num_qpts = GetCommonPlaneSegmentRule( pen_enfrc_options.common_plane_quadrature_order,
+                                                        segment_rule_wts, segment_rule_coords );
+        const RealT x0 = xVert[0];
+        const RealT y0 = xVert[1];
+        const RealT x1 = xVert[2];
+        const RealT y1 = xVert[3];
+        const RealT length = magnitude( x1 - x0, y1 - y0 );
+
+        for ( int qp = 0; qp < num_qpts; ++qp ) {
+          const RealT s = segment_rule_coords[qp];
```
srwopschall, line 700: Minor: consider for clarity adding comment about what this mapping is.

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+          RealT vel_q1[max_dim] = { 0., 0., 0. };
+          RealT vel_q2[max_dim] = { 0., 0., 0. };
+
+          const bool mapped_face1 =
+              EvalLinearEdgeAtProjectedPoint( actual_xf1, x_q, overlapNormal, x_qf1, phi_q1, use_rate ? dim : 0,
+                                              use_rate ? &actual_vf1[0] : nullptr, use_rate ? vel_q1 : nullptr );
+          const bool mapped_face2 =
+              EvalLinearEdgeAtProjectedPoint( actual_xf2, x_q, overlapNormal, x_qf2, phi_q2, use_rate ? dim : 0,
+                                              use_rate ? &actual_vf2[0] : nullptr, use_rate ? vel_q2 : nullptr );
+          if ( !mapped_face1 || !mapped_face2 ) {
```
srwopschall, line 717: Same comment. Add debug statement here? Is this bad if we get in this if-block?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -289,6 +581,172 @@ int ApplyNormal<COMMON_PLANE, PENALTY>( CouplingScheme* cs )
...
+
+            RealT local_gap = ( x_qf1[0] - x_qf2[0] ) * overlapNormal[0] + ( x_qf1[1] - x_qf2[1] ) * overlapNormal[1] +
+                              ( x_qf1[2] - x_qf2[2] ) * overlapNormal[2];
+            if ( local_gap > gap_tol ) {
+              continue;
+            }
+
+            has_contact = true;
+
+            RealT local_pressure = local_gap * penalty_stiff_per_area;
```
srwopschall, line 669: Can you clarify whether this routine before computed the pressure? As a sanity check is this line inside ComputeGapRatePressure? Shouldn't this only compute the pressure as done on line 675? Or did the original version of this routine compute it all?

src/tribol/physics/CommonPlane.cpp
```diff
@@ -506,17 +922,171 @@ int ApplyTangential<COMMON_PLANE, PENALTY, VISCOUS_TANGENTIAL>( CouplingScheme*
...
+    RealT phi2[max_nodes_per_face];
     initRealArray( phi1, numNodesPerFace1, 0. );
     initRealArray( phi2, numNodesPerFace2, 0. );
 
     ////////////////////////////////////////////////////////////////////////
     // Integration of contact integrals: integral of shape functions over //
     // contact overlap patch                                              //
     ////////////////////////////////////////////////////////////////////////
-    EvalWeakFormIntegral<COMMON_PLANE, SINGLE_POINT>( cntctElem, phi1, phi2 );
+    EvalWeakFormIntegralCommonPlane( cntctElem, pen_enfrc_options.common_plane_rule,
```
srwopschall, line 1088: I think here we need to clarify which set of face coordinates cntctElem has and how they are used in this routine and whether that conflicts with multipoint integration using current configuration physical coords.

src/tribol/physics/Mortar.cpp
```diff
@@ -28,8 +28,8 @@ void ComputeMortarWeights( SurfaceContactElem& elem )
   // instantiate integration object
   IntegPts integ;
 
-  // Debug: leave code in for now to call Gauss quadrature on triangle rule
-  GaussPolyIntTri( elem, integ, 3 );
+  // Mortar retains the historical triangle rule for now.
+  GaussPolyIntTri( elem, integ, 3, TRI_RULE_LEGACY );
 
   // call Taylor-Wingate-Bos integation rule. NOTE: this is not
```
srwopschall, line 34: Since we are touching integration rules, the TWB integration rule never worked and was experimental at the time. It is not consistent with the underlying finite element basis functions. Let's consider removing.

srwopschall (commented):
@ebchin - a lot of comments, but only a handful of things that really need to be ironed out. Please review. In the meantime I need to look at the entirety of CommonPlane.cpp and verify the code and when/where average face coordinates are used vs. current configuration and possibly warped face coordinates. I want to be sure we don't have any inconsistencies.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added a comment describing the planar/warped quad configuration and the local penetration.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Replaced the raw coupling scheme id and quadrature order with named constants.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Renamed EdgeLocalContactResult to EdgeLocalContactForceResult for consistency with WarpedQuadForceResult.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added a comment describing the fully overlapping tilted-edge contact configuration.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Replaced the raw coupling scheme id and quadrature order with named constants.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added a high-level comment describing what the multipoint quad execution test covers.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added the planar full-overlap, uniform-penetration mesh configuration to the test comment.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added a high-level comment describing the triangular contact-surface test.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added the planar full-overlap, uniform-penetration mesh configuration to the test comment.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Tightened the warped-quad gap checks to the expected computed common-plane gap.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Tightened the warped-quad force check to the expected multi-point force magnitude and kept the zero net z-force check.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Tightened the edge-local-contact gap, total force, and imbalance checks to expected numeric values.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Added a comment explaining that the imbalance is the difference between the mesh-1 nodal force norms.

src/tests/tribol_common_plane_penalty.cpp
ebchin: Removed the soft threshold and replaced it with explicit expected imbalance values.

src/tests/tribol_enforcement_options.cpp
ebchin: Added a high-level comment describing what the stored integration-options test checks.

src/tests/tribol_enforcement_options.cpp
ebchin: Cleaned up the test by using named constants and the same names in the expectations.

src/tests/tribol_iso_integ.cpp
ebchin: Added comments defining the normalized reference-triangle moment helper.

PR #210, opened 2026-06-01
Optional dt calc

ebchin:
Allows a formulation to report it doesn't have a dt calc, then skips dt calculation.

### DIFF
--- a/src/tribol/mesh/CouplingScheme.cpp
+++ b/src/tribol/mesh/CouplingScheme.cpp
@@ -1035,7 +1035,9 @@ int CouplingScheme::apply( int cycle, RealT t, RealT& dt )
     }
     m_formulation_impl->updateNodalGaps();
     m_formulation_impl->updateNodalForces();
-    dt = m_formulation_impl->computeTimeStep();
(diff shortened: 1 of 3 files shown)


src/tribol/mesh/CouplingScheme.cpp
```diff
@@ -1035,7 +1035,9 @@ int CouplingScheme::apply( int cycle, RealT t, RealT& dt )
     }
     m_formulation_impl->updateNodalGaps();
     m_formulation_impl->updateNodalForces();
-    dt = m_formulation_impl->computeTimeStep();
+    if ( m_formulation_impl->hasTimeStepCalculation() ) {
```
srwopschall, line 1038: I'm assuming this doesn't affect any of the common plane timestep stuff?

srwopschall (approved):

src/tribol/mesh/CouplingScheme.cpp
ebchin: No, just the stuff that uses the `ContactFormulation` class

PR #213, opened 2026-06-12
Use reference coords for element thickness calc

ebchin:


src/tribol/mesh/MfemData.cpp
```diff
@@ -969,6 +982,8 @@ void MfemMeshData::ComputeElementThicknesses()
     quad_val[0] = h;
   }
 
+  parent_mesh.SetNodes( saved_nodes );
```
srwopschall: Should this have been there before? Or how is this related to the element thickness?

src/tribol/mesh/MfemData.cpp
```diff
@@ -918,6 +918,19 @@ void MfemMeshData::SetLORFactor( int lor_factor )
 
 void MfemMeshData::ComputeElementThicknesses()
```
srwopschall, line 921: Is there a way to add a unit test for this with two different coordinate configurations to see if this is doing the right thing? It looks fine, but would be nice to test.

src/tribol/mesh/MfemData.cpp
ebchin: The new code changes the coordinates stored on the mesh to be the reference coords, so it computes the right detJ. This line restores the original coordinates stored in the mesh.

src/tribol/mesh/MfemData.cpp
ebchin: test added

srwopschall (approved):

PR #215, opened 2026-07-06
TPL Update

chapman39:
- Takes host configs from https://github.com/llnl/smith/pull/1600
- also has commit from https://github.com/llnl/Tribol/pull/214, but separated for clarity
- requires https://github.com/llnl/Tribol/pull/207 in order for all tests to pass!

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -12,9 +12,9 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/develop' }}
 
 env:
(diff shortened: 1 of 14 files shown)


chapman39:
just to clarify some things,

this branch was branched from the hubcast PR. however, i reverted the TPL build changes in that PR. so github thinks the contents of `task/chapman39/tpl-update-2026-07-06` merged, however that is actually not the case. i have a 2nd tpl build PR that should be merged instead: https://github.com/llnl/Tribol/pull/218

PR #216, opened 2026-07-06
hubcast

chapman39:
integrate CI with hubcast. the pipelines will only run via hubcast after this PR merges, which is why the gitlab CI is not appearing ATM.

### DIFF
--- a/.github/hubcast.yml
+++ b/.github/hubcast.yml
@@ -1,0 +1,28 @@
+Repo:
+  # Required: organization or user that owns the repo on the destination forge
+  dest_org: tribol
+
+  # Required: name of the destination repository
+  dest_name: tribol
+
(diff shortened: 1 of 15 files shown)


chapman39:
@lc-hubcast help

.github/hubcast.yml
```diff
@@ -0,0 +1,28 @@
+Repo:
+  # Required: organization or user that owns the repo on the destination forge
+  dest_org: tribol
+
+  # Required: name of the destination repository
+  dest_name: tribol
```
ebchin, line 6: I think this should be tribol-ci?

chapman39: my understand is it needs to align with https://lc.llnl.gov/gitlab/tribol/tribol/ 

the first tribol is the `organization`, the 2nd is the repository `name`

ebchin (approved):
Thanks @chapman39

Issue #229, opened 2026-08-03
`Intersection2DPolygon` can return clockwise overlap vertices for nearly coincident triangles


ebchin:
Running the [contact patch example in Smith](https://github.com/llnl/smith/blob/develop/src/smith/physics/tests/contact_patch.cpp) uncovered fragility in determining CCW orientation in `Intersection2DPolygon()` for the general intersection path.  When intersecting polygons are nearly overlapping, but not overlapping enough to trigger the all A in B or all B in A paths, then [`PolyReorderConvex()`](https://github.com/llnl/Tribol/blob/develop/src/tribol/geom/GeomUtilities.hpp#L1467) can fail. Nearly overlapping polygons have some vertices that are nearly identical, which aren't merged until a subsequent call to [`CheckPolySegs()`](https://github.com/llnl/Tribol/blob/develop/src/tribol/geom/GeomUtilities.hpp#L1484).  When calling `PolyReorderConvex()` with nearly overlapping vertices, the vertices potentially do not lie on the boundary of a convex polygon, breaking the assumptions of `PolyReorderConvex()`.

Issue reproducer:

```
#include "tribol/geom/GeomUtilities.hpp"

#include <iostream>
#include <iomanip>

static double signedArea(const double* x, const double* y, int n)
{
  double a = 0.0;
  for (int i = 0; i < n; ++i) {
    const int j = (i + 1) % n;
    a += x[i] * y[j] - y[i] * x[j];
  }
  return 0.5 * a;
}

int main()
{
  // Nonmortar projected triangle from failing contact_patch case.
  const double xA[3] = {
      -1.7720926303008128e-01,
       1.7720926303008130e-01,
       7.6327832942979648e-17
  };
  const double yA[3] = {
      -5.9069754343360363e-02,
      -5.9069754343360349e-02,
       1.1813950868672080e-01
  };

  // Mortar projected triangle after ElemReverse() in ComputeMortarForceEnzyme.
  const double xB[3] = {
       1.7720926303008075e-01,
      -8.3266726846886605e-17,
      -1.7720926303008128e-01
  };
  const double yB[3] = {
      -5.9069754343361050e-02,
       1.1813950868672063e-01,
      -5.9069754343360363e-02
  };

  double polyX[8] = {};
  double polyY[8] = {};
  int numPolyVert = 0;
  double area = 0.0;

  auto err = tribol::Intersection2DPolygon(
      xA, yA, 3,
      xB, yB, 3,
[...]

PR #230, opened 2026-08-03
unify length scales in Intersection2DPolygon

ebchin:
Fixes the cause of #229 for the geometry identified. Tolerances in `Intersection2DPolygon()` are used inconsistently. Specifically for the geometry identified, this causes a tolerance for determining if one polygon is not contained another to not be triggered, but a similar tolerance checking if the points should be collapsed to be triggered down the line.  Between the two checks, `PolyReorderConvex()` is called, but since the geometry contains non-collapsed points, it is not convex, causing the routine to fail.

This PR includes 2 fixes:
- makes the use of tolerances in `Intersection2DPolygon()` consistent
- merge near-identical points before calling `PolyReorderConvex()`

### DIFF
--- a/src/tests/tribol_enzyme_element_mortar.cpp
+++ b/src/tests/tribol_enzyme_element_mortar.cpp
@@ -968,7 +968,10 @@ TEST_F( EnzymeElementMortarTest, NoOverlap )
                              1.0,  1.0,  1.0,  1.0 };
   // clang-format on
 
-  FDCheck( x1, x2, n1, p1, x1_stencil, x2_stencil, 4, 1.0, 1.0e-7 );
+  constexpr int num_nodes = 4;
+  constexpr double check_scale = 1.0;
+  constexpr double len_collapse_ratio = 1.0e-7;
+  FDCheck( x1, x2, n1, p1, x1_stencil, x2_stencil, num_nodes, check_scale, len_collapse_ratio );
 }
 
 }  // namespace tribol
--- a/src/tests/tribol_enzyme_element_mortar.cpp
+++ b/src/tests/tribol_enzyme_element_mortar.cpp
@@ -49,7 +49,8 @@ class EnzymeElementMortarTest : public testing::Test {
    * @param x2_stencil Stencil coordinate directions of the second face
    */
   void FDCheck( double* x1, double* x2, double* n1, double* p1, const double* x1_stencil = nullptr,
-                const double* x2_stencil = nullptr, int num_nodes = 4, double check_scale = 1.0 )
+                const double* x2_stencil = nullptr, int num_nodes = 4, double check_scale = 1.0,
+                double len_collapse_ratio = 1.0e-8 )
   {
     constexpr int max_num_disp_dofs = 12;
     constexpr int max_num_pres_dofs = 4;
@@ -85,7 +86,8 @@ class EnzymeElementMortarTest : public testing::Test {
     double df2dp1[max_num_disp_dofs * max_num_pres_dofs] = { 0.0 };
 
     tribol::ComputeMortarJacobianEnzyme( x1, n1, p1, f1, df1dx1, df1dx2, df1dn1, df1dp1, g1, dg1dx1, dg1dx2, dg1dn1,
-                                         num_nodes, x2, f2, df2dx1, df2dx2, df2dn1, df2dp1, num_nodes );
+                                         num_nodes, x2, f2, df2dx1, df2dx2, df2dn1, df2dp1, num_nodes,
+                                         len_collapse_ratio );
 
     // Finite difference derivative of the force on the first face w.r.t. the coordinates of the first face
     double df1dx1_fd[max_num_disp_dofs * max_num_disp_dofs] = { 0.0 };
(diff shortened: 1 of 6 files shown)


src/tribol/geom/GeomUtilities.hpp
```diff
@@ -1531,8 +1576,9 @@ TRIBOL_HOST_DEVICE inline FaceGeomException Intersection2DPolygon(
  * \param [in] xB array of local x coordinates of polygon B
  * \param [in] yB array of local y coordinates of polygon B
  * \param [in] numVertexB number of vertices in polygon B
- * \param [in] posTol position tolerance to collapse segment-segment intersection points
- * \param [in] lenTol length tolerance to collapse short intersection edges
+ * \param [in] lengthTol nondimensional length tolerance used directly for point-in-face barycentric checks and
```
white238: Congrats on the longest parameter description I've seen!

white238 (approved):
Was this what was exposed by the submodule PR?

src/tribol/geom/GeomUtilities.hpp
ebchin: Brevity is the soul of technical miscommunication. --Codex, probably

src/tests/tribol_enzyme_element_mortar.cpp
```diff
@@ -966,7 +968,7 @@ TEST_F( EnzymeElementMortarTest, NoOverlap )
                              1.0,  1.0,  1.0,  1.0 };
   // clang-format on
 
-  FDCheck( x1, x2, n1, p1, x1_stencil, x2_stencil );
+  FDCheck( x1, x2, n1, p1, x1_stencil, x2_stencil, 4, 1.0, 1.0e-7 );
```
srwopschall: 4, 1.0, and 1.0e-7 is a little cryptic. Can we name them?

src/tribol/geom/CompGeom.hpp
```diff
@@ -1322,12 +1319,10 @@ TRIBOL_HOST_DEVICE inline FaceGeomException MortarPlanePair::computeOverlap3D( c
   ElemReverse( X2, Y2, m2.numberOfNodesPerElement() );
 
   // compute intersection polygon and area.
-  RealT pos_tol = this->m_params.len_collapse_ratio *
-                  axom::utilities::max( m1.getFaceRadius()[element_id1], m2.getFaceRadius()[element_id2] );
-  RealT len_tol = pos_tol;
+  RealT length_tol = this->m_params.len_collapse_ratio;
```
srwopschall: Before this len_collapse_ratio multiplied the maximum face radius to make the tolerance proportional to the mesh. Can we restore this? If this is 1.e-8 we need to sort of normalize it against a mesh dimension.

src/tribol/geom/CompGeom.hpp
```diff
@@ -2511,12 +2503,10 @@ TRIBOL_HOST_DEVICE inline FaceGeomException CommonPlanePair::projectPointsAndCom
   }
 
   // call intersection routine to get intersecting polygon
-  RealT pos_tol = this->m_params.len_collapse_ratio *
-                  axom::utilities::max( m1.getFaceRadius()[element_id1], m2.getFaceRadius()[element_id2] );
-  RealT len_tol = pos_tol;
+  RealT length_tol = this->m_params.len_collapse_ratio;
```
srwopschall: Same comment. We need to apply this ratio to a mesh dimension like before.

src/tribol/geom/GeomUtilities.hpp
```diff
@@ -1139,8 +1154,9 @@ enum class OverlapVertexType
...
```
srwopschall: Consider more concise explanation?

src/tribol/geom/GeomUtilities.hpp
```diff
@@ -1219,6 +1235,11 @@ TRIBOL_HOST_DEVICE inline FaceGeomException Intersection2DPolygon(
   VertexAvgCentroid( xA, yA, nullptr, numVertexA, xCA, yCA, zC );
   VertexAvgCentroid( xB, yB, nullptr, numVertexB, xCB, yCB, zC );
 
+  const RealT lengthScaleA = Polygon2DLengthScale( xA, yA, numVertexA );
+  const RealT lengthScaleB = Polygon2DLengthScale( xB, yB, numVertexB );
+  const RealT faceLengthScale = ( lengthScaleA > lengthScaleB ) ? lengthScaleA : lengthScaleB;
+  const RealT physicalLengthTol = lengthTol * faceLengthScale;
```
srwopschall: Just clarify regarding previous comments...a lengthTol is set to the ratio on the params struct, which is misleading, but then it is the case that later on this is actually still being treated as a ratio, where here for instance, we apply it to the physical length scale of the elements an get the actual physical length tolerance?

srwopschall (approved):

src/tests/tribol_enzyme_element_mortar.cpp
ebchin: renamed

src/tribol/geom/CompGeom.hpp
ebchin: Now it's scaled internally in the method. Renamed length_tol_ratio so it's clear.

src/tribol/geom/CompGeom.hpp
ebchin: See above

src/tribol/geom/GeomUtilities.hpp
ebchin: fixed

src/tribol/geom/GeomUtilities.hpp
ebchin: Right, it sends a ratio in now. The issue is there is a tolerance based on a barycentric coordinate calculation in this routine, which wants a nondimensional length. I renamed to lengthTolRatio.

PR #232, opened 2026-08-04
Simplify `PolyReorderConvex()` to use angles from the centroid

ebchin:
This is another robustness fix related to #229.  The previous routine assumes a convex hull and breaks when that assumption is violated. The new routine is simpler (just get the vertex average and compute angles around it) and can work with some star-convex geometries (provided the angular scan from the vertex average "sees" all of the boundary of the polygon).

### DIFF
--- a/src/tests/tribol_comp_geom_3d.cpp
+++ b/src/tests/tribol_comp_geom_3d.cpp
@@ -2494,6 +2494,52 @@ TEST_F( CompGeomTest, poly_reorder_convex_1 )
   EXPECT_EQ( y[4], 0.5 );
 }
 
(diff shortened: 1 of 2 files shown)


srwopschall (commented):
Can/should we add a unit test for this? I like having the CG routines unit tested.

ebchin:
> Can/should we add a unit test for this? I like having the CG routines unit tested.

There's already a unit test on line 2400 of tribol_comp_geom_3d.cpp. I added another one that tests a triangle, a quad, and a star convex polygon.

srwopschall (approved):

PR #236, opened 2026-08-10
Task/ebchin/build warnings changes

ebchin:
- Turn on warnings as errors so we (hopefully) don't have to do this again
- Try to fix the MPI and OpenMP logic in the SetupMFEM.cmake file so we don't have to list them as dependencies in executables that depend on MFEM
- Use axom::numeric_limits<RealT>::max() instead of std::numeric_limits<RealT>::max() for device support
- Make spatialDimension() an int


cmake/Options.cmake
```diff
@@ -19,6 +19,7 @@ endif()
 
 option(TRIBOL_USE_SINGLE_PRECISION "Use single-precision floating point" OFF)
 option(TRIBOL_USE_64BIT_INDEXTYPE "Use 64-bit index type" OFF)
+option(TRIBOL_ENABLE_WARNINGS_AS_ERRORS "Treat compiler warnings as errors in Tribol" ON)
```
white238: You don't want this to be a CMake dependent option?

https://github.com/llnl/axom/blob/2957850014f3899363fcfddc7f83c8ab853ebe44/src/cmake/AxomOptions.cmake#L48

cmake/thirdparty/SetupMFEM.cmake
```diff
@@ -140,17 +154,34 @@ else()
...
         get_filename_component(_mpi_lib_dir ${_first_mpi_lib} DIRECTORY)
         target_link_directories(mfem INTERFACE ${_mpi_lib_dir})
-        target_link_libraries(mfem INTERFACE mpi)
+        target_link_libraries(mfem INTERFACE blt::mpi)
     endif()
     # End Tribol edit
 
 endif()
 
+if(MFEM_BUILT_WITH_CMAKE AND _mfem_uses_mpi)
```
white238: You should unify this with 165. Either move this up to the line 25 block or have the logic in 157 down here as well.. basically deduplicate this somehow.

CMakeLists.txt
```diff
@@ -79,6 +79,9 @@ if ("${PROJECT_SOURCE_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
     set(BLT_REQUIRED_CLANGFORMAT_VERSION  "19" CACHE STRING "")
 endif()
 
+set(ENABLE_ALL_WARNINGS ON CACHE BOOL "")
+set(ENABLE_WARNINGS_AS_ERRORS ON CACHE BOOL "")
```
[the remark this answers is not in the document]
white238: This is a good suggestion. It makes it follow the parent project if tribol is being included in another project.

cmake/thirdparty/SetupMFEM.cmake
ebchin: fixed

src/tribol/geom/CompGeom.hpp
```diff
@@ -2465,6 +2466,9 @@ TRIBOL_HOST_DEVICE inline FaceGeomException CommonPlanePair::projectPointsAndCom
                   << "input number of vertices to match number of nodes per element." );
     }
   }
+#else
+  TRIBOL_UNUSED_VAR( m1 );
```
white238, line 2470: any reason you dont use `[[maybe_unused]]`?

white238: This could be a follow up PR across tribol

white238 (approved):

src/tribol/geom/CompGeom.hpp
ebchin: No good reason. I agree -- worth cleaning up in a follow-on PR.

PR #241, opened 2026-09-02
EnergyMortar quad pt: ensure tangent is calculated when gap = 0

ebchin:


src/tribol/physics/EnergyMortar.hpp
```diff
@@ -34,6 +34,7 @@ struct QuadraturePointPenaltyData {
   static constexpr int num_force_dofs = dim * max_nodes_per_elem * pair_size;  ///< Pair coordinate degrees of freedom.
   static constexpr int num_stiffness_entries = num_force_dofs * num_force_dofs;  ///< Flattened stiffness size.
 
+  bool has_active_qp{ false };                            ///< True when any quadrature-point gap is nonpositive.
```
srwopschall, line 37: Just or clarification note that nonpositive is interpenetration. Is this sort of like an "in_contact" boolean, that if any quadrature point for a given quad rule is nonpositive then it indicates that there is contact for this face-pair?

src/tribol/physics/EnergyMortar.cpp
```diff
@@ -515,12 +515,18 @@ TRIBOL_ENZYME_INLINE double qp_penalty_kernel_qp_energy( double xiA, double w, c
   const double dy = x1[1] - x2[1];
   const double gn = -( dx * nB[0] + dy * nB[1] );
   const double gap = gn * eta;
+  const bool is_active = gap <= 0.0;
```
srwopschall, line 519: Perhaps consider clarifying the equality here. Usually we only consider a pair active if the gap is less than zero per tribol convention. Let's discuss this.

src/tribol/physics/EnergyMortar.cpp
```diff
@@ -515,12 +515,18 @@ TRIBOL_ENZYME_INLINE double qp_penalty_kernel_qp_energy( double xiA, double w, c
...
 
-  return gap < 0.0 ? 0.5 * penalty * gap * gap * w * J : 0.0;
+  *has_active_qp = *has_active_qp || is_active;
```
srwopschall: Looks like two booleans controlling the same thing? Or if has_active_qp is returned then should it just be *has_active_qp = is_active, since is_active is set based on the gap, which seems like the controlling thing?

src/tribol/physics/EnergyMortarAdapter.cpp
```diff
@@ -525,7 +525,7 @@ void QuadraturePoint<Adapter>::updateNodalForces()
     const auto qp_data =
         adapter->evaluator_->compute_quadrature_point_penalty_data( flipped_pair, mesh1_view, mesh2_view );
 
-    if ( qp_data.energy == 0.0 ) {
+    if ( !qp_data.has_active_qp ) {
```
srwopschall, line 528: Let's discuss the bugfix part of this. Looks like we continued if the energy was equal to zero, which could imply that the gap is zero. Now, we consider active qps for gap == 0, right?

srwopschall (commented):
Let's chat real fast about this one, but otherwise all looks good.

src/tribol/physics/EnergyMortar.cpp
ebchin: The change here is to match nodal enforcement for EnergyMortar. We want gap <= 0 to be active so the stiffness is calculated when the constraint is exactly enforced, even though the force and energy are zero.

This is relevant mostly at the initial mesh configuration.

More generally, it's probably worth it to put gap definition in a single class so we define it once and re-use it where it's needed.

src/tribol/physics/EnergyMortar.cpp
ebchin: `is_active` is good for this particular quad pt. `has_active_qp` is the value for all the quad pts over the CouplingScheme.

src/tribol/physics/EnergyMortar.hpp
ebchin: Correct

src/tribol/physics/EnergyMortarAdapter.cpp
ebchin: Yeah, that's it

srwopschall (approved):

PR #246, opened 2026-09-21
fix tribol version in config header

chapman39:
small pr that sets the `TRIBOL_VERSION_` cmake variables before the `tribol/config.hpp.in` header gets configured. this will allow the config header to properly define the version macros.


ebchin (approved):
Thanks @chapman39

srwopschall (approved):

